PR #34, opened 2026-02-19
adding spec-to-run roadmap

bbe64:
## Description

This PR updates the `ROADMAP.md` to include a new major security and compliance feature: **Trusted Spec-to-Run Pipeline**.

As Barbacane targets enterprise and highly regulated environments, users need cryptographically verifiable guarantees that the running Data Plane is executing the exact configuration/spec they approved, with zero drift.

### The feature covers:
1. **Build-time Fingerprinting**: Injecting a cryptographic hash (and Git commit / S3 object-id) into the `.bca` artifact during compilation.
2. **Runtime Verification**: Exposing an API endpoint on the Data Plane to query the current config fingerprint.
3. **Control Plane Drift Detection**: The Control Plane will periodically verify the running instances against the expected source of truth and alert on anomalies.
4. **Supply Chain Security**: Exporting this provenance data into the SBOM when running as an OCI container.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -146,6 +146,16 @@ Near-term items ready to be picked up:
 | Homebrew formula | macOS package manager support | P3 |
 | APT/RPM packages | Linux package manager support | P3 |
 
+### Security & Config provenance: Trusted Spec-to-Run Pipeline
+
+To guarantee that the running gateway is executing the exact specification intended, we will implement a verifiable spec-to-run trust chain. This ensures no unauthorized configuration drift can occur between the source repository and the production edge.
+
+- [ ] **Artifact fingerprinting:** Automatically calculate and embed a cryptographic hash (along with optional metadata like Git commit SHA or S3 Object ID) into the `.bca` artifact during the `barbacane compile` step.
+- [ ] **Provenance API endpoint:** Add a dedicated, authenticated endpoint to the Data Plane to interrogate the currently loaded configuration fingerprint.
+- [ ] **Control Plane drift detection:** Implement a background worker in `barbacane-control` that periodically polls connected Data Planes to ensure the running artifact hash matches the authorized version, triggering alerts upon drift.
+- [ ] **OCI / SBOM integration:** Surface the specification fingerprint in the software bill of materials (SBOM) and container labels when packaging the Data Plane as an OCI image, completing the end-to-end supply chain verification.
+
+
 ---
 
 ## Technical Debt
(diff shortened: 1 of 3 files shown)


ndreno:
Nice ADR! One question on the hash scope:

The ADR says the `spec_hash` is a "SHA-256 hash of all input configuration files" — does this cover only the spec YAML/JSON, or does it also include the plugin WASM binaries bundled into the `.bca`?

A swapped-out plugin binary would alter gateway behavior just as much as a changed spec, so it seems like the hash should cover everything that goes into the artifact. Worth clarifying in the ADR either way.

bbe64:
> Nice ADR! One question on the hash scope:
> 
> The ADR says the `spec_hash` is a "SHA-256 hash of all input configuration files" — does this cover only the spec YAML/JSON, or does it also include the plugin WASM binaries bundled into the `.bca`?
> 
> A swapped-out plugin binary would alter gateway behavior just as much as a changed spec, so it seems like the hash should cover everything that goes into the artifact. Worth clarifying in the ADR either way.

Great catch! You are absolutely right. 

Securing the declarative configuration is useless if a malicious actor can swap out a WASM plugin before compilation. The WASM binaries are executable code and their integrity is just as critical as the routing rules. 

I'll update the ADR to explicitly state that the `spec_hash` must be a combined hash (e.g., a Merkle tree root or a hash of a sorted manifest) covering both the YAML/JSON specs **and** all referenced WASM binaries bundled into the `.bca` artifact. 

Thanks for pointing this out, making the update now!

ndreno:
Another thought: the ADR mentions exposing `/_admin/provenance` on a "dedicated admin port/interface", but Barbacane doesn't currently have a separate admin listener — the data plane only binds a single port for user traffic.

Introducing a dedicated admin port is a non-trivial prerequisite (new CLI flags, separate TLS config, bind address handling, etc.). It might be worth either:

- Calling it out explicitly as a dependency/prerequisite in the ADR, or
- Deferring it to its own ADR (e.g. "ADR-0022: Admin API listener") and referencing it here

Just so the scope is clear when someone picks this up for implementation.

ndreno:
One more on the drift detection design:

The ADR describes the Control Plane polling Data Planes via HTTP (`GET /_admin/provenance`), but today the communication works the other way around — it's the **Data Plane that connects to the Control Plane** over a WebSocket (`ws://control-plane/ws/data-plane`). The Control Plane has no way to initiate HTTP requests back to Data Plane instances.

So rather than introducing a new HTTP polling mechanism, drift detection could piggyback on the existing WebSocket channel: the Data Plane would report its artifact hash when it connects (and periodically thereafter), and the Control Plane would compare it against the expected value on its side.

This avoids the need for a reverse connectivity path and keeps the existing network topology intact (Data Planes can sit behind NATs/firewalls without issue).

bbe64:
@ndreno pushed an update according to your remarks.

ndreno (approved):

PR #37, opened 2026-02-25
Adr/0020 response transformers

NathanPrats:
## Summary

Add `response-transformer` middleware plugin — the second half of ADR-0020 (request/response transformers). Provides declarative response transformations before client delivery: status code mapping, header transforms, and JSON body transforms.

## Related Issues

Implements `response-transformer` from ADR-0020.

## Changes

- New `response-transformer` plugin with status code mapping, header (add/set/remove/rename), and body (add/remove/rename via JSON Pointer RFC 6901) transformations
- 32 unit tests covering all transform operations, edge cases, and config deserialization
- 6 integration tests with dedicated test fixture (`response-transformer.yaml`)
- 6 manual test scenarios in `tests/api.http`
- Documentation: CHANGELOG, README plugin table, ROADMAP, middlewares guide, extensions reference

## Type of Change

- [x] New feature (non-breaking change that adds functionality)
- [x] Documentation update
- [x] Test update

## Testing

- [x] Unit tests pass (`cargo test` — 32 passing)
- [x] Integration tests pass
- [x] Manual testing performed

<details>
<summary>Test details</summary>

**Unit tests (32):**
- Status mapping: basic, unmapped passthrough, multiple mappings, empty map
- Headers: add, add overwrites, set (no overwrite), remove, remove case-insensitive, rename, rename nonexistent, transformation order, combined add+remove
- Body: add simple, add nested, add overwrites, remove, remove nested, rename, rename nested, transformation order, non-JSON passthrough, empty body, empty string, nonexistent pointer, array pointer
- Integration: combined (status+headers+body), empty config passthrough, on_request passthrough, deserialization defaults, deserialization full, status-only config

**Integration tests (6):**
- Status code mapping (200 → 201)
- Multiple status mappings
- Header transformations (add/remove/rename)
- Body transformations (add/remove/rename)
- Combined (all transforms together)
- Passthrough baseline

</details>

## Checklist

- [x] My code follows the project's style guidelines (`cargo fmt` and `cargo clippy`)
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix/feature works
- [x] New and existing tests pass locally
- [x] I have updated documentation as needed
- [x] My commits are signed off (DCO)

## Additional Notes

- **No variable interpolation in v1** — keeping scope minimal. Can be layered in later when needed.
- **Status mapping uses `BTreeMap<String, u16>`** — YAML/JSON deserializes integer keys as strings, so we match on stringified status codes.
- Follows the single-file `lib.rs` pattern established by Nicolas's polished request-transformer.
- WASM binary is gitignored (CI builds it).


ndreno (commented):
Great work on this — the structure is clean, the test coverage is solid, and the transformation ordering is well-documented. I went through the code and addressed a few things directly in `pr/37` (commit `39c2452`). Here's what changed and why:

**Bug: rename used two different pointer APIs** (`lib.rs`)

The rename loop was reading with `serde_json::Value::pointer()` (stdlib) but deleting with `jsonptr::Pointer`. These two disagree on RFC 6901 escape sequences (`~0`/`~1`). A field named `foo~bar` referenced as `/foo~0bar` would resolve correctly for the read but fail silently on the delete, corrupting the body. Fixed by switching the read to `old_ptr.resolve(&json)` so both operations go through `jsonptr` consistently. The same latent bug exists in `request-transformer` — I'll follow up there separately.

**Bug: rename with identical source and destination deleted the field**

`rename: { /field: /field }` would assign the value to itself, then delete it — field gone. Added an `if old_ptr == new_ptr { continue; }` guard. Added `test_body_rename_same_pointer` as a regression test.

**Schema: no range constraint on status mapping target values**

`200: 999` or `200: 0` would be accepted silently. Added `"minimum": 100, "maximum": 599` to the `additionalProperties` in `config-schema.json`.

**Minor: `or_insert_with(|| value.clone())` → `or_insert(value.clone())`**

The closure wasn't needed for a trivially cheap clone.

**Minor: refactored `if let / else { panic! }` to `let-else`** in `test_on_request_passthrough` — reduces nesting, same semantics.

**New tests (36 total, was 32)**
- `test_headers_remove_and_readd_same_key` — same key in both `remove` and `add`
- `test_body_rename_same_pointer` — regression for the no-op rename fix
- `test_body_rename_array_element_field` — rename inside an array element
- `test_body_remove_array_element_field` — remove inside an array element

One thing I consciously did **not** change: `body.add` keeping all values as strings. I initially flagged this but after reading `request-transformer` more carefully — it only coerces via `to_json_value` for interpolated variables (`$header.*`), not for literal config values. Since `response-transformer` has no variable interpolation the string-only behaviour is correct and consistent.

PR #39, opened 2026-03-04
feat: AI gateway ADRs (streaming, AI proxy, MCP)

ndreno:
## Summary

- **ADR-0023: WASM Plugin Streaming Support** — New `host_http_stream` host function for SSE/chunked streaming with buffered post-processing. Unblocks AI gateway and future streaming use cases.
- **ADR-0024: AI Gateway Plugin** — `ai-proxy` dispatcher (OpenAI, Anthropic, Ollama) with provider fallback, contract-first API versioning, and composable middlewares (`ai-token-limit`, `ai-cost-tracker`, `ai-prompt-guard`, `ai-response-guard`).
- **ADR-0025: MCP Server Support** — Automatic MCP tool generation from OpenAPI specs as a native data plane feature. Includes `x-barbacane-mcp` config following the global/operation override pattern.

All three ADRs are **Proposed** status — no code changes.

## Test plan

- [x] Review ADR-0023 for streaming design feasibility
- [x] Review ADR-0024 for AI gateway plugin architecture and provider coverage
- [x] Review ADR-0025 for MCP server design choices
- [x] Verify cross-references between ADRs are consistent

### DIFF
--- a/adr/0023-wasm-plugin-streaming.md
+++ b/adr/0023-wasm-plugin-streaming.md
@@ -8,7 +8,6 @@
 The current dispatch contract requires plugins to return a fully buffered `Response` (status + headers + `Option<String>` body). This works for typical API proxying but breaks down for use cases that require streaming responses to clients:
 
 - **LLM completions** — Chat APIs (OpenAI, Anthropic) stream tokens via SSE, often taking 10–60 seconds. Buffering the entire response before sending it to the client defeats the purpose of streaming and creates unacceptable latency.
-- **Large file proxying** — S3 objects, report downloads, export APIs.
 - **Event streams** — Server-Sent Events, webhook relays.
 
 Barbacane's middleware pipeline (SPEC-002) currently assumes a complete response is available before the on_response chain runs:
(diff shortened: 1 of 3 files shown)


adr/0023-wasm-plugin-streaming.md
```diff
@@ -0,0 +1,104 @@
...
+
+**Status:** Proposed
+**Date:** 2026-03-03
+
+## Context
+
+The current dispatch contract requires plugins to return a fully buffered `Response` (status + headers + `Option<String>` body). This works for typical API proxying but breaks down for use cases that require streaming responses to clients:
+
+- **LLM completions** — Chat APIs (OpenAI, Anthropic) stream tokens via SSE, often taking 10–60 seconds. Buffering the entire response before sending it to the client defeats the purpose of streaming and creates unacceptable latency.
+- **Large file proxying** — S3 objects, report downloads, export APIs.
```
bbe64: You mentioned "Large file proxying — S3 objects, report downloads..." as a use case. If the host buffers the entire response body internally while streaming to the client, proxying a 5GB S3 object will buffer 5GB in the host's memory, which will inevitably lead to OOM crashes under load.

Recommendation:
We should probably remove "Large file proxying" from the immediate use cases, or explicitly state in the Trade-offs section that this MVP implementation is strictly safe for text/LLM payloads (typically <100KB). We should also explicitly track the "fire-and-forget" non-buffering mode as a hard requirement before we can support large file proxying.

adr/0024-ai-gateway-plugin.md
```diff
@@ -0,0 +1,422 @@
...
+
+This leverages the middleware pipeline: each concern is independently ordered, configured, and optional. Users who only need proxying skip the middlewares. Users who need cost tracking add it. Users who need guardrails compose them in the chain.
+
+### Provider Support (MVP)
+
+| Provider | Format | Translation | Auth |
+|----------|--------|-------------|------|
+| OpenAI | OpenAI API | Passthrough | `Authorization: Bearer <key>` |
+| Anthropic | Messages API | Request + response translation | `x-api-key: <key>` |
+| Ollama | OpenAI-compatible | Passthrough | None (local) |
```
bbe64, line 42: Should we support vLLM in the MVP ? https://vllm.ai/

bbe64 (dismissed):
Let's ride the AI train.

adr/0023-wasm-plugin-streaming.md
ndreno: Good catch. We dug into this and here's where we landed:

**Request side is already covered** — the data plane enforces `--max-body-size` (default 1 MB) with early Content-Length rejection before body read, plus the `request-size-limit` middleware plugin for per-operation control, and the WASM sandbox caps plugin memory at 16 MB.

**Response side is the real concern.** A `response-size-limit` middleware would help as a safety net for unexpectedly large upstream responses (say 50 MB instead of 10 KB), but it wouldn't solve the multi-GB case — the host has already buffered the full response in memory before `on_response` runs, so OOM happens before the middleware can reject it.

**Updated the ADR:**
- Removed "Large file proxying" from the immediate use cases
- Expanded the Trade-offs section to explicitly state this MVP is safe for text/LLM payloads (<100KB) only
- Noted that large file proxying requires the future non-buffering "fire-and-forget" stream mode
- Referenced existing request-side limits as defense in depth

See 9694832.

adr/0024-ai-gateway-plugin.md
ndreno: vLLM exposes an OpenAI-compatible API, so it already works via `provider: openai` + `base_url: http://vllm:8000` — same story as Ollama, TGI, LocalAI, etc.

Rather than adding a dedicated provider entry for each compatible server, added a note clarifying that any OpenAI-compatible inference server works out of the box with the `openai` provider and a custom `base_url`. No adapter needed.

See 9a2db33.

bbe64 (approved):

PR #41, opened 2026-03-06
feat: streaming dispatch via host_http_stream (ADR-0023)

ndreno:
## Summary

- Implements `host_http_stream` in the WASM runtime so dispatcher plugins can stream SSE/chunked upstream responses directly to clients without buffering the full body in the gateway process
- Adds `streamed_response()` sentinel to the plugin SDK, documenting the streaming contract for dispatcher authors
- Wires up the streaming path in the data plane: `dispatch_wasm_plugin_inner` now races `stream_rx.recv()` against the blocking WASM task via `tokio::select!`, returning a `StreamBody` (boxed) for streaming or running `on_response` inline for buffered responses
- Restructures `barbacane-test` from a 4200-line monolithic `mod tests` block into 15 focused integration test files under `tests/`, each covering a distinct feature area
- Adds `tests/fixture-plugins/streaming-echo` — a minimal WASM dispatcher that exercises the new streaming path — compiled automatically by `build.rs`

## Test plan

- [x] `cargo test --workspace --exclude barbacane-test` — unit tests pass
- [x] `cargo test -p barbacane-test streaming` — streaming integration tests pass (requires `wasm32-unknown-unknown` target and Docker services)
- [x] `cargo clippy --all-targets` — zero errors
- [x] `cargo fmt --all` — no diffs


tests/fixture-plugins/streaming-echo/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "streaming-echo"
+version = "0.1.0"
+edition = "2021"
+description = "Fixture plugin for streaming integration tests — proxies to upstream via host_http_stream"
+license = "MIT"
```
klaude, line 6: This may need to be the Apache 2.0 license to match the base project. Though I see all package licenses are MIT, so it may be better to adopt MIT in the overall project. I don't know if that's something to address in this PR, but it's something to look out for.

crates/barbacane-wasm/src/http_client.rs
```diff
@@ -586,6 +649,90 @@ mod tests {
         assert_eq!(tls.ca, Some(PathBuf::from("/etc/certs/ca.crt")));
     }
 
+    // ── stream_raw validation ─────────────────────────────────────────────────
```
klaude, line 652: Is it possible to test HTTP timeouts, connection failures, other failures, and a successful request?

ndreno: Added tests for connection failure, timeout (using a TCP listener that never accepts), and a successful end-to-end request (minimal HTTP server via tokio TcpListener). See c2b2cd9.

tests/fixture-plugins/streaming-echo/Cargo.toml
ndreno: Good catch — all workspace crates use MIT today. I'll leave it as-is for this PR and we can address the license alignment project-wide separately.

PR #42, opened 2026-03-08
feat: add vacuum ruleset for Barbacane spec extensions

ndreno:
## Summary

- Publish a vacuum ruleset at `docs.barbacane.dev/rulesets/barbacane.yaml` that validates `x-barbacane-dispatch` and `x-barbacane-middlewares` extensions against plugin JSON schemas — catching config errors, missing dispatchers, and security misconfigurations at lint time
- 17 rules across 5 categories: dispatch validation, middleware validation, extension hygiene, upstream/secrets, and auth safety
- 7 custom JS functions for schema validation, duplicate detection, secret ref format, and auth opt-out checks
- Plugin `config-schema.json` files copied from source at docs build time (single source of truth)
- Documentation page with quick start, rule reference, and CI integration guide

## Customers usage

```yaml
# .vacuum.yml
extends:
  - - https://docs.barbacane.dev/rulesets/barbacane.yaml
    - recommended
```

## Test plan

- [x] 11/11 ruleset tests passing (`docs/rulesets/tests/run-tests.sh`)
- [x] Valid specs produce 0 barbacane violations (6 fixtures tested)
- [x] Invalid specs correctly trigger violations: missing dispatch, unknown plugins, bad configs, plaintext HTTP, malformed secret refs, unknown extensions
- [x] `cargo fmt --all` clean
- [x] `cargo clippy --all-targets` clean
- [x] `cargo test --workspace --exclude barbacane-test` passing


docs/guide/vacuum.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Quick Start
+
+### 1. Install vacuum
+
+```bash
+# macOS
+brew install daveshanley/vacuum/vacuum
+
+# or download from https://github.com/daveshanley/vacuum/releases
```
klaude: [Vacuum's installation docs](https://quobix.com/vacuum/installing/) have a lot of installation alternatives to homebrew.

docs/guide/vacuum.md
```diff
@@ -0,0 +1,124 @@
...
+```
+
+### 2. Create a `.vacuum.yml` in your project
```
klaude, line 18: Generally speaking, the `.yaml` extension is preferred for YAML files in modern projects. It also helps resolve the 1 in 1214141423341234 chance that someone confuses it with [YML](https://fdik.org/yml/).

docs/rulesets/barbacane.yaml
```diff
@@ -0,0 +1,227 @@
+extends: [[spectral:oas, recommended]]
```
klaude: Consider using `vacuum:oas` instead of `spectral:oas` to pick up on the vacuum author's latest supported rules.

docs/rulesets/functions/barbacane-auth-opt-out.js
```diff
@@ -0,0 +1,75 @@
...
+  if (!Array.isArray(globalMiddlewares)) return results;
+
+  const hasGlobalAuth = globalMiddlewares.some(
+    (m) => m && AUTH_PLUGINS.has(m.name)
+  );
+  if (!hasGlobalAuth) return results;
+
+  // Scan all operations
+  const paths = input.paths;
+  if (!paths || typeof paths !== "object") return results;
```
klaude: Across the custom functions `results` is returned before anything is pushed to it. Consider returning an empty array in these cases and declare `results` closer to where it's used.

docs/rulesets/functions/barbacane-no-plaintext-upstream.js
```diff
@@ -0,0 +1,31 @@
+// Warns when http-upstream dispatcher uses plaintext HTTP instead of HTTPS.
```
klaude: Technically speaking, the warning is thrown in the ruleset file that uses this rule.

docs/rulesets/functions/barbacane-valid-secret-refs.js
```diff
@@ -0,0 +1,43 @@
...
+
+function checkValues(obj, results) {
+  if (!obj || typeof obj !== "object") return;
+
+  for (const [key, value] of Object.entries(obj)) {
+    if (typeof value === "string") {
+      const isSecretRef = SECRET_PREFIXES.some((p) => value.startsWith(p));
+      if (!isSecretRef) continue;
+
+      if (value.startsWith("env://") && !ENV_REF_RE.test(value)) {
```
klaude: You may not need to call `startsWith()` since the regular expression check right after will do the same thing.

docs/rulesets/functions/barbacane-validate-dispatch-config.js
```diff
@@ -0,0 +1,172 @@
...
+    case "integer":
+      return typeof value === "number" && Number.isInteger(value);
+    case "boolean":
+      return typeof value === "boolean";
+    case "object":
+      return typeof value === "object" && !Array.isArray(value);
+    case "array":
+      return Array.isArray(value);
+    default:
+      return true;
```
klaude: Be wary of unintended side effects from defaulting to a valid object type if `expectedType` is unknown.

docs/rulesets/functions/barbacane-validate-middleware-config.js
```diff
@@ -0,0 +1,305 @@
...
+          message: `Config field "${key}" for middleware "${pluginName}" must be <= ${prop.maximum}.`,
+        });
+      }
+    }
+  }
+
+  return results;
+}
+
+function checkType(value, expectedType) {
```
klaude, line 287: This duplicates the same function in `docs/rulesets/functions/barbacane-validate-dispatch-config.js`. It may be useful to declare a utility library for common code. Some of the env:// and file:// handling could go in there too.

docs/rulesets/tests/run-tests.sh
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/env bash
+# Test runner for the Barbacane vacuum ruleset.
+# Usage: ./docs/rulesets/tests/run-tests.sh
+#
+# Requires: vacuum (https://quobix.com/vacuum/)
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
```
klaude: Consider adding the `-P` option to `pwd` to resolve symlinks. That may matter on some systems.

docs/rulesets/tests/run-tests.sh
```diff
@@ -0,0 +1,83 @@
...
+ROOT_DIR="$(cd "$SCRIPT_DIR/../../.." && pwd)"
+RULESET="$SCRIPT_DIR/../barbacane.yaml"
+FUNCTIONS="$SCRIPT_DIR/../functions"
```
klaude: Rename this to `FUNCTIONS_DIR` for consistency with the other directory variable names.

docs/rulesets/tests/run-tests.sh
```diff
@@ -0,0 +1,83 @@
...
+PASS=0
+FAIL=0
+
+# Count barbacane-specific violations using spectral-report JSON output.
+count_barbacane_violations() {
+  local spec="$1"
+  vacuum spectral-report -r "$RULESET" --functions "$FUNCTIONS" -o "$spec" 2>/dev/null \
```
klaude: Try the vacuum [`report`](https://quobix.com/vacuum/commands/report/) command instead of `spectral-report`. 

Also, try the `-n` option. You may be able to use plain ol' `sed`, `grep`, and `wc` to parse the results then.

docs/rulesets/tests/run-tests.sh
```diff
@@ -0,0 +1,83 @@
...
+    | sed -n '/^\[/,$p' \
+    | python3 -c "
```
klaude: There may be a clever `jq` one-liner that does this, removing the need for python.

docs/rulesets/tests/run-tests.sh
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/env bash
+# Test runner for the Barbacane vacuum ruleset.
```
klaude, line 2: Something like [bats](https://bats-core.readthedocs.io/en/stable/index.html) may be nice to remove some of the maual scaffolding and get you test run artifacts for reports, if that's a thing you want.

docs/rulesets/barbacane.yaml
ndreno: Good point on `vacuum:oas`, but going a step further and dropping the `extends` entirely. The ruleset is scoped exclusively to `x-barbacane-*` extensions — none of the rules depend on OAS base rules. Pulling in `spectral:oas`/`vacuum:oas` would impose ~40 general OAS rules on every user who extends this ruleset, potentially conflicting with whatever OAS baseline they already have.

The intent is that users compose their own stack:
```yaml
extends:
  - [vacuum:oas, recommended]
  - [https://docs.barbacane.dev/rulesets/barbacane.yaml, recommended]
```
Done in the latest commit.

docs/rulesets/functions/barbacane-validate-middleware-config.js
ndreno: Good catch on the `checkType` duplication. Unfortunately, vacuum loads each function file independently via its `goja` JS runtime — there's no `require()`/`import` support between function files, so a shared utility module isn't directly possible in this model.

The `checkType` function is small enough that the duplication is manageable, and keeping it inline in each file avoids a loading-order dependency that doesn't exist. I've fixed the `default: return true` → `return false` issue you flagged in both files though.

docs/guide/vacuum.md
ndreno: Noted preference for `.yaml`. However, vacuum uses `.vacuum.yml` as its default config file name — users can run `vacuum lint` without `-r` only if the file is named `.vacuum.yml`. Switching to `.vacuum.yaml` in the docs would require users to always pass `-r .vacuum.yaml` explicitly, which is worse UX. Keeping `.yml` to match vacuum's convention.

docs/rulesets/tests/run-tests.sh
ndreno: bats looks nice for the longer term. Out of scope for this PR but noted as a potential improvement if the test suite grows.

docs/rulesets/tests/run-tests.sh
ndreno: `vacuum report` generates an HTML file, which is harder to parse programmatically. `spectral-report` gives JSON output that is straightforward to query — and with the jq one-liner replacing Python (done in the latest commit), parsing is now cleaner. Keeping `spectral-report`.

docs/guide/vacuum.md
klaude: I 100% agree. It's all good!

docs/rulesets/functions/barbacane-validate-middleware-config.js
klaude: I think I was speaking from experience writing Spectral rules in TypeScript which transpiled into a single JavaScript file to include in my rulesets. That is totally overkill for this case though!

docs/rulesets/tests/run-tests.sh
klaude: Eek, `-P` should be an option to `pwd` instead of `cd`.

ndreno: Done — moved `-P` from `cd` to `pwd` in 904974b. All review comments should be addressed now. Could you give it another look and approve if everything looks good? Thanks!

klaude: Nah mate. Looks good!

klaude (approved):

PR #44, opened 2026-03-09
docs: add ADR-0027 for OCI artifact distribution

thomassaison:
## Summary

- Adds ADR-0027 defining the strategy for distributing `.bca` artifacts via OCI registries using ORAS
- Specifies the `application/vnd.barbacane.artifact.v1+bca` media type
- Documents the init container pattern for Kubernetes (Helm chart) artifact provisioning
- Establishes a minimum `oras` version (`v1.2.0`) and digest-pinning requirement for security
- Sketches a future `barbacane compile --push` convenience command

## Test plan

- [x] Review ADR content and wording
- [x] Confirm media type naming convention aligns with OCI artifact guidance

### DIFF
--- a/adr/0026-oci-artifact-distribution.md
+++ b/adr/0026-oci-artifact-distribution.md
@@ -33,6 +33,8 @@ oras push ghcr.io/my-org/my-api:1.0.0 \
 
 **In Kubernetes (Helm chart):** The data plane Deployment includes an init container running the official `oras` image that pulls the artifact into a shared `emptyDir` volume. The main container mounts the same volume.
 
+> **Note:** ADR-0019 deferred Helm chart creation until deployment patterns stabilized. This ADR describes the init container pattern that informed the Helm chart design; the chart itself now lives in `barbacane-helm-chart/`.
+
 ```yaml
 initContainers:
   - name: artifact-fetcher
@@ -42,21 +44,46 @@ initContainers:
       - $(ARTIFACT_REF)
       - --output
       - /artifact
+      - --registry-config
+      - /registry-config/.dockerconfigjson
+    env:
+      - name: ARTIFACT_REF
+        value: ghcr.io/my-org/my-api@sha256:abc123...  # set via Helm values
+    securityContext:
+      runAsNonRoot: true
+      runAsUser: 65532
+      runAsGroup: 65532
+      allowPrivilegeEscalation: false
+      readOnlyRootFilesystem: true
+      capabilities:
+        drop: ["ALL"]
     volumeMounts:
       - name: artifact
         mountPath: /artifact
+      - name: registry-config
+        mountPath: /registry-config
+        readOnly: true
 containers:
   - name: barbacane
     args: ["serve", "--artifact", "/artifact/api.bca"]
+    securityContext:
+      runAsNonRoot: true
+      runAsUser: 65532
+      runAsGroup: 65532
     volumeMounts:
       - name: artifact
         mountPath: /artifact
+        readOnly: true
 volumes:
   - name: artifact
-    emptyDir: {}
+    emptyDir:
+      sizeLimit: 100Mi  # tune via Helm values for largest expected artifact
+  - name: registry-config
+    secret:
+      secretName: artifact-registry-credentials  # Docker config secret for oras pull
 ```
 
-Registry credentials are provided via `imagePullSecrets`, reusing the existing Kubernetes mechanism.
+**Registry credentials:** `imagePullSecrets` on the Pod spec covers pulling the `oras` init container image itself. For artifact pulls, the `oras` CLI needs credentials injected separately — here via a mounted Docker config secret (`--registry-config`). Both the init container and the main container run as UID 65532 (`nonroot`) per ADR-0019, ensuring the artifact file is readable by both.
 
 ### Security
 
@@ -64,7 +91,9 @@ The init container image must be pinned to a **minimum version of `v1.2.0`** in 
 - Verified digest-based pulls (preventing tag mutation attacks)
 - Improved TLS certificate validation
 
-In production, the Helm chart defaults to a pinned digest (`ghcr.io/oras-project/oras@sha256:...`) rather than a floating tag, with the tag kept as a human-readable comment. Users overriding the image are responsible for maintaining this guarantee.
+In production, the Helm chart defaults to a pinned digest (`ghcr.io/oras-project/oras@sha256:...`) rather than a floating tag, with the tag kept as a human-readable comment. The chart maintainers are responsible for updating the pinned digest when upgrading the `oras` image version. Users overriding the image are responsible for maintaining this guarantee in their own values.
+
+Digest-pinned artifact pulls also provide an integrity guarantee that complements ADR-0021's provenance requirements: the pull is content-addressed, ensuring the artifact bytes match what was originally pushed.
 
 ### Future: `barbacane compile --push`
 
@@ -82,7 +111,7 @@ As a convenience, a future `barbacane compile --push <registry-ref>` subcommand 
 - **Versioning:** Artifact versions are explicit OCI tags, not filesystem paths
 - **Reuse:** Same registry, same auth, same tooling as container images
 - **Auditability:** Registry logs artifact pulls alongside image pulls
-- **Provenance:** Artifact hash and metadata can be attached as OCI annotations, tying back to the artifact fingerprinting work (ADR-0021)
+- **Provenance:** Digest-pinned pulls are content-addressed, providing integrity guarantees that tie directly into ADR-0021's provenance chain. Artifact hash and metadata can be attached as OCI annotations
 
 ### Harder
 
--- a/adr/0026-oci-artifact-distribution.md
+++ b/adr/0026-oci-artifact-distribution.md
@@ -1,0 +1,103 @@
+# ADR-0026: OCI Artifact Distribution for .bca Artifacts
+
+**Status:** Proposed
+**Date:** 2026-03-09
+
+## Context
+
+The Barbacane data plane requires a compiled `.bca` artifact at startup (`serve --artifact /config/api.bca`). In Kubernetes deployments, this artifact must be provisioned into the pod before the gateway starts.
+
+Several distribution approaches exist:
+
+- **ConfigMap/Secret:** Limited by 1MB etcd object size; not suitable for large artifacts
+- **PersistentVolume:** Requires external CI/CD to populate the volume; ties artifact lifecycle to cluster infrastructure
+- **Init container (container image):** Works but repurposes container image semantics — mixing application images with data artifacts
+- **OCI artifacts (ORAS):** Stores the `.bca` as a typed, versioned artifact in any OCI-compatible registry, cleanly separated from container images
+
+Barbacane already uses OCI registries for container images (ADR-0019). Using the same registry infrastructure for `.bca` artifacts gives users a single place to manage all Barbacane-related distribution assets.
+
+## Decision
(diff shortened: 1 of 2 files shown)


adr/0026-oci-artifact-distribution.md
```diff
@@ -0,0 +1,103 @@
...
+    args: ["serve", "--artifact", "/artifact/api.bca"]
+    volumeMounts:
+      - name: artifact
+        mountPath: /artifact
+volumes:
+  - name: artifact
+    emptyDir: {}
+```
+
+Registry credentials are provided via `imagePullSecrets`, reusing the existing Kubernetes mechanism.
```
ndreno: **[Blocking] `imagePullSecrets` does not cover runtime `oras pull` credentials**

`imagePullSecrets` is a Kubernetes mechanism for pulling *container images*. It controls whether the kubelet can pull `ghcr.io/oras-project/oras` — but it has no effect on the `oras pull` command that runs *inside* that init container at runtime.

For a private artifact registry, the `oras` CLI needs credentials injected explicitly. Common patterns:

**Option A — Mount a Docker config secret:**
```yaml
initContainers:
  - name: artifact-fetcher
    image: ghcr.io/oras-project/oras@sha256:...
    args:
      - pull
      - $(ARTIFACT_REF)
      - --output
      - /artifact
      - --registry-config
      - /registry-config/.dockerconfigjson
    volumeMounts:
      - name: artifact
        mountPath: /artifact
      - name: registry-config
        mountPath: /registry-config
        readOnly: true
volumes:
  - name: registry-config
    secret:
      secretName: my-registry-pull-secret
```

**Option B — Environment variables:**
```yaml
env:
  - name: ORAS_USERNAME
    valueFrom:
      secretKeyRef:
        name: registry-credentials
        key: username
  - name: ORAS_PASSWORD
    valueFrom:
      secretKeyRef:
        name: registry-credentials
        key: password
```
(then pass `--username $(ORAS_USERNAME) --password $(ORAS_PASSWORD)` in `args`)

Note: `imagePullSecrets` should still be present on the Pod spec to pull the `oras` init container image itself if it lives in a private registry — but it is separate from authenticating artifact pulls.

adr/0027-oci-artifact-distribution.md
```diff
@@ -0,0 +1,103 @@
...
+
+**In Kubernetes (Helm chart):** The data plane Deployment includes an init container running the official `oras` image that pulls the artifact into a shared `emptyDir` volume. The main container mounts the same volume.
+
+```yaml
+initContainers:
+  - name: artifact-fetcher
+    image: ghcr.io/oras-project/oras:v1.2.0  # minimum version, see Security below
+    args:
+      - pull
+      - $(ARTIFACT_REF)
```
ndreno, line 44: `$(ARTIFACT_REF)` is undefined in this snippet. Kubernetes environment variable substitution requires a corresponding `env:` block on the init container. Either add it, or replace with a concrete placeholder to keep the example self-contained (and to reinforce the digest-pinning guidance from the Security section below).

Suggested addition to the init container spec:
```yaml
env:
  - name: ARTIFACT_REF
    value: ghcr.io/my-org/my-api@sha256:abc123...  # set via Helm values
```

Using a digest reference here is also better than a tag — it makes the YAML snippet consistent with the Security section's own recommendation.

adr/0027-oci-artifact-distribution.md
```diff
@@ -0,0 +1,103 @@
...
+  --artifact-type application/vnd.barbacane.artifact.v1+bca \
+  api.bca:application/vnd.barbacane.artifact.v1+bca
+```
+
+**In Kubernetes (Helm chart):** The data plane Deployment includes an init container running the official `oras` image that pulls the artifact into a shared `emptyDir` volume. The main container mounts the same volume.
+
+```yaml
+initContainers:
+  - name: artifact-fetcher
+    image: ghcr.io/oras-project/oras:v1.2.0  # minimum version, see Security below
```
ndreno, line 41: **Missing security context on the init container.** Per ADR-0019, all containers should run as non-root. The init container needs an explicit `securityContext`, otherwise it defaults to running as root (which the `oras` base image may do).

Add:
```yaml
    securityContext:
      runAsNonRoot: true
      runAsUser: 65532        # match the nonroot user from ADR-0019
      runAsGroup: 65532
      allowPrivilegeEscalation: false
      readOnlyRootFilesystem: true
      capabilities:
        drop: ["ALL"]
```

One interaction to be aware of: if the init container writes to `/artifact` as UID 65532, the main `barbacane` container must run as the same UID (or have group read permissions on the volume) to read the file. This is typically fine since both use `nonroot:nonroot`, but worth making explicit in the ADR so operators don't encounter a silent permission failure.

adr/0026-oci-artifact-distribution.md
```diff
@@ -0,0 +1,103 @@
...
+        mountPath: /artifact
+containers:
+  - name: barbacane
+    args: ["serve", "--artifact", "/artifact/api.bca"]
+    volumeMounts:
+      - name: artifact
+        mountPath: /artifact
+volumes:
+  - name: artifact
+    emptyDir: {}
```
ndreno: An unbounded `emptyDir` can exhaust node disk space for large `.bca` artifacts, potentially causing disk pressure that affects other pods on the same node.

Recommend adding a `sizeLimit` tunable via Helm values:
```yaml
volumes:
  - name: artifact
    emptyDir:
      sizeLimit: 100Mi  # tune to your largest expected artifact; set via .Values.artifact.sizeLimit
```

This is a Kubernetes best practice for any emptyDir used as a data transfer volume.

adr/0027-oci-artifact-distribution.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+**Pushing an artifact:**
+```bash
+oras push ghcr.io/my-org/my-api:1.0.0 \
+  --artifact-type application/vnd.barbacane.artifact.v1+bca \
+  api.bca:application/vnd.barbacane.artifact.v1+bca
+```
+
+**In Kubernetes (Helm chart):** The data plane Deployment includes an init container running the official `oras` image that pulls the artifact into a shared `emptyDir` volume. The main container mounts the same volume.
```
ndreno, line 34: **Tension with ADR-0019's explicit Helm chart deferral.**

ADR-0019 under 'What We're NOT Doing' states: *"Helm charts: Defer until deployment patterns stabilize."* This ADR references 'the Helm chart' multiple times as if it already exists or is a settled decision.

These two ADRs are in mild contradiction. Please resolve it with one of:

1. Acknowledge the tension here explicitly: *"This ADR describes the deployment pattern that, once validated, will inform the Helm chart design deferred in ADR-0019. The YAML snippet is illustrative — a Helm chart is not in scope for this iteration."*
2. Or update the status of ADR-0019's Helm deferral if the team has decided to move forward with it now.

As written, an implementer reading both ADRs will not know whether a Helm chart exists, is planned, or is still deferred.

ndreno (commented):
Good architecture overall — the ORAS init-container pattern is a clean fit, and the rationale for rejecting ConfigMap/PV/image-repurposing is well argued. A few issues to fix before this lands, one of them blocking.

**Blocking**
- The `imagePullSecrets` section is technically incorrect for private registries — see inline comment on line 59.

**Medium**
- Init container YAML is missing a security context (non-root, drop caps, read-only root FS)
- `emptyDir` should have a `sizeLimit`
- `$(ARTIFACT_REF)` is undefined in the YAML snippet
- The Helm chart references create a tension with ADR-0019's explicit deferral of Helm charts

**Low**
- Digest pinning maintenance ownership is unspecified
- The integrity guarantee from digest-pinned pulls should be explicitly connected to ADR-0021's provenance requirements

Details inline below.

ndreno:
**ADR numbering conflict** — PR #43 ([feat/adr-0026-websocket-proxy](https://github.com/barbacane-dev/barbacane/pull/43)) is also open and uses ADR-0026 for the WebSocket transparent proxy. Since that PR was opened first, this OCI artifact distribution ADR should be renumbered to **ADR-0027** (both the file name `adr/0027-oci-artifact-distribution.md` and all internal references).

adr/0026-oci-artifact-distribution.md
thomassaison: Fixed. Replaced `imagePullSecrets` with a mounted Docker config secret (`--registry-config`) for `oras pull` credentials. Added a paragraph clarifying the distinction: `imagePullSecrets` covers pulling the oras container image, while the mounted secret covers artifact pulls.

adr/0027-oci-artifact-distribution.md
thomassaison: Fixed. Added an `env:` block with a digest-pinned reference (`ghcr.io/my-org/my-api@sha256:abc123...`) so the snippet is self-contained and consistent with the Security section.

adr/0027-oci-artifact-distribution.md
thomassaison: Fixed. Added full `securityContext` (non-root UID 65532, drop all caps, read-only root FS) on the init container, and matching `runAsUser`/`runAsGroup` on the main container so both share the same UID for volume access.

adr/0026-oci-artifact-distribution.md
thomassaison: Fixed. Added `sizeLimit: 100Mi` with a comment to tune via Helm values.

adr/0027-oci-artifact-distribution.md
thomassaison: Acknowledged the tension with a note: ADR-0019 deferred Helm charts until deployment patterns stabilized; this ADR describes the init container pattern that informed the chart design, and the chart now lives in `barbacane-helm-chart/`.

ndreno (dismissed):
All issues addressed — clean round of fixes. The blocking `imagePullSecrets` / `oras pull` credentials distinction is now clearly explained, the YAML snippet is self-contained with a digest-pinned env var, the init container has a proper security context, and the `emptyDir` is bounded. The two low-priority items (digest pinning ownership and the ADR-0021 provenance link) are both covered in the Security section and Consequences.

The Helm chart tension with ADR-0019 is resolved with the right framing — pattern first, chart follows.

ndreno:
**Checklist item: Confirm media type naming convention aligns with OCI artifact guidance**

Looked into this — the current media type `application/vnd.barbacane.artifact.v1+bca` has a naming issue.

The OCI image spec requires `artifactType` values to comply with **RFC 6838**, which reserves the `+` structured syntax suffix for *registered* formats only (`+json`, `+xml`, `+gzip`, `+tar`, etc.). Using `+bca` as a suffix is technically non-conformant since `.bca` is a proprietary binary format with no IANA-registered suffix.

For comparison, projects in the OCI ecosystem that get this right:
- `application/vnd.cncf.helm.chart.v1.tar+gzip` — `+gzip` is a registered suffix
- `application/vnd.dev.cosign.artifact.sig.v1+json` — `+json` is a registered suffix

**Recommended fix:** drop the suffix and use dot-separated notation:
- `application/vnd.barbacane.artifact.v1` (minimal)
- `application/vnd.barbacane.bca.v1` (makes the format explicit without misusing `+`)

Checklist item cannot be checked off as-is — the media type needs to be corrected in the ADR before this lands.

thomassaison:
**Re: media type RFC 6838 compliance**

Fixed. Renamed to `application/vnd.barbacane.bca.v1` — drops the non-conformant `+bca` structured syntax suffix per RFC 6838. The format is explicit in the type name without misusing the `+` suffix.

ndreno (approved):

PR #45, opened 2026-03-12
feat: add ws-upstream WebSocket transparent proxy dispatcher

ndreno:
## Summary

- Add `ws-upstream` dispatcher plugin for transparent WebSocket proxying (ADR-0026)
- 4-layer implementation: WASM plugin → `host_ws_upgrade` host function → `ws_client` module → data plane 101 sentinel + bidirectional frame relay
- Plugin validates upgrade headers, builds upstream URL with path param substitution and query forwarding, filters hop-by-hop headers, adds X-Forwarded-For/Host

## Changes

**Plugin** (`plugins/ws-upstream/`): dispatcher with config schema, 24 unit tests  
**Host function** (`barbacane-wasm`): `host_ws_upgrade` connects upstream via tokio-tungstenite, stores stream in `PluginState`  
**Data plane** (`barbacane`): detects WS upgrade before body consumption, handles 101 sentinel, spawns `relay_websocket` for bidirectional frame forwarding  
**Capability**: `ws_upgrade` registered in manifest system  
**Tests**: 3 integration tests + fixture spec  
**Docs**: ws-upstream sections in dispatchers guide and extensions reference

## Test plan

- [x] 24 unit tests in `plugins/ws-upstream/` (config, upgrade validation, dispatch, path substitution, error responses)
- [x] 3 integration tests: spec compiles, rejects non-WS requests, handles unavailable upstream
- [x] `cargo clippy --all-targets` passes
- [x] `cargo test --workspace --exclude barbacane-test` passes
- [ ] Manual test with a real WebSocket echo server

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -577,6 +577,7 @@ dependencies = [
  "tempfile",
  "thiserror 2.0.18",
  "tokio",
+ "tokio-tungstenite",
  "toml",
  "tracing",
  "uuid",
--- a/crates/barbacane-test/tests/plugins.rs
+++ b/crates/barbacane-test/tests/plugins.rs
@@ -1,4 +1,4 @@
-//! Integration tests for utility middleware plugins — CORS, correlation-id, request-size-limit, IP restriction, NATS, Kafka, HTTP log, request-transformer, response-transformer.
+//! Integration tests for utility middleware plugins — CORS, correlation-id, request-size-limit, IP restriction, NATS, Kafka, HTTP log, request-transformer, response-transformer, ws-upstream.
 //!
 //! Run with: `cargo test -p barbacane-test`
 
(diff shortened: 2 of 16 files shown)


NathanPrats:
# Review: feat/ws-upstream — Critical Issue

## Missing `Sec-WebSocket-Accept` in the 101 Response

**File:** `crates/barbacane/src/main.rs`
**Severity:** Critical

### Problem

The 101 Switching Protocols response sent back to the client is built without a `Sec-WebSocket-Accept` header:

```rust
let response = Response::builder()
    .status(StatusCode::SWITCHING_PROTOCOLS)
    .header("upgrade", "websocket")
    .header("connection", "Upgrade")
    .body(BoxBody::new(Full::new(Bytes::new())))
    .expect("valid 101 response");
```

RFC 6455 §4.2.2 mandates that the server's 101 response includes a `Sec-WebSocket-Accept` header derived from the client's `Sec-WebSocket-Key`. Any compliant WebSocket client — browsers, most WS libraries — will reject the handshake if this header is absent or invalid.

This is why tests may still pass: `wscat` and lenient HTTP clients do not enforce the header. But browser-based consumers and strict clients will silently fail to upgrade.

### Fix

Extract the `Sec-WebSocket-Key` from the incoming request headers, compute the accept value per RFC 6455, and include it in the 101 response.

```rust
use base64::{engine::general_purpose::STANDARD, Engine};
use sha1::{Digest, Sha1};

fn compute_ws_accept(key: &str) -> String {
    const WS_GUID: &str = "258EAFA5-E914-47DA-95CA-C5AB0DC85B11";
    let mut hasher = Sha1::new();
    hasher.update(key.as_bytes());
    hasher.update(WS_GUID.as_bytes());
    STANDARD.encode(hasher.finalize())
}
```

Then in the 101 response builder:

```rust
let ws_key = headers
    .get("sec-websocket-key")
    .and_then(|v| v.to_str().ok())
    .unwrap_or_default();

let accept = compute_ws_accept(ws_key);

let response = Response::builder()
    .status(StatusCode::SWITCHING_PROTOCOLS)
    .header("upgrade", "websocket")
    .header("connection", "Upgrade")
    .header("sec-websocket-accept", accept)
    .body(BoxBody::new(Full::new(Bytes::new())))
    .expect("valid 101 response");
```

### Required dependencies

`sha1` and `base64` are already indirect dependencies via `jsonwebtoken` / `ring`. Add them explicitly to `crates/barbacane/Cargo.toml`:

```toml
base64 = "0.22"
sha1 = "0.10"
```

ndreno:
Fixed in 1652712 — the 101 response now includes `Sec-WebSocket-Accept` computed from the client's `Sec-WebSocket-Key` per RFC 6455 §4.2.2. Uses `tungstenite::handshake::derive_accept_key` (already a transitive dep), so no new crates needed.

NathanPrats (approved):

PR #102, opened 2026-07-06
fix(wasm): WebSocket SSRF guard (completes WA-9) + security-hardening docs

ndreno:
## What

`host_ws_upgrade` connected to any `ws://`/`wss://` URL a plugin supplied **with no egress check**, so a plugin could reach internal / loopback / cloud-metadata services over WebSocket — bypassing the SSRF guard that already covers HTTP and Kafka/NATS.

## Fix

`ws_client::connect_upstream` now resolves the upstream host and refuses internal targets via the shared `guard_external_host`, unless `BARBACANE_ALLOW_INTERNAL_EGRESS` is set. The egress policy is threaded from the `Gateway`.

## Scope

Completes the broker/WS egress-guard portion of **WA-9** (private #3). Kafka and NATS already guard + time out + bound their connection caches; the HTTP outbound-header denylist landed in #99. After this, #3's only remaining item is the optional WA-5 bounded worker pool (per-call timeouts + the epoch-deadline refresh already mitigate the "pin a worker" concern).

## Tests

- Unit: `connect_upstream` blocks a loopback target when egress is disallowed.
- Existing ws-upstream integration tests (egress allowed) still reach loopback mocks — all 6 green locally.
- Workspace clippy + `cargo check --workspace --lib --bins` clean.


ndreno:
Added a docs commit bringing the guide/reference up to date for the whole hardening series (#98-#102): ALLOW_INTERNAL_EGRESS now documents WebSocket + DNS-rebinding safety, the writeOnly/E1070 secret convention is documented for plugin authors and operators, the vacuum function-download list is de-drifted, and the CHANGELOG covers all of #98-#102.

Issue #138, opened 2026-09-08
WASM runtime: per-request instantiation, engine-global plugin limits, and a seven-function WASI surface


ndreno:
Three properties of the WASM plugin runtime that cost something on every request today, and that also gate any future heavyweight plugin. Surfaced while assessing WAF integration options in ADR-0031, where they were prerequisites for the sidecar and TinyGo routes. **The route that ADR chose is a native pipeline stage, so none of this blocks the WAF work.** Filing it separately so it can be scheduled on its own merits.

They are independent of each other and can land in any order.

---

### P1. A fresh WASM instance is created and `init()`-ed on every request, per middleware

`InstancePool::get_instance` builds a new `PluginInstance` and calls `init(config)` each time it is invoked:

- `crates/barbacane-wasm/src/pool.rs:223` (`get_instance`)
- `crates/barbacane-wasm/src/pool.rs:260` (`instance.init(&config_json)`)

and it is invoked per request, per middleware:

- `crates/barbacane/src/main.rs:1576` (middleware chain construction)
- `crates/barbacane/src/main.rs:1866` (dispatch)

Modules are AOT-compiled and cached, so this is instantiation plus guest-side `init`, not compilation. For the current plugin set that is cheap but not free: a request through five middlewares pays five instantiations and five `init` calls, and `init` runs guest code that parses the plugin's config JSON.

The name `InstancePool` suggests instances are pooled. They are not; the type caches modules and configs.

**What to do:** reuse instances keyed on `(plugin, config)`, which is what `InstanceKey` already models, so `init` runs once per key rather than once per request. Needs a decision on guest state: a reused instance carries whatever the guest left in its linear memory, so either instances are reset between requests or plugins gain a documented contract that `on_request` must not depend on prior state. That contract question is the real work here, not the pooling.

**Unmeasured.** I have not benchmarked this, so I am not claiming a number. A benchmark comparing the current path against a reused instance across the existing plugin set would size it, and is worth doing before the change rather than after.

### P2. Plugin resource limits are engine-global, so every plugin gets the largest plugin's ceiling

`PluginLimits` defaults to 16 MB of linear memory, 100 ms wall clock and 100 M fuel (`crates/barbacane-wasm/src/limits.rs:28`), and the data plane installs one set for the whole engine:

- `crates/barbacane/src/main.rs:734`

[...]

PR #139, opened 2026-09-08
docs: add ADR-0031 (WAF integration, ModSecurity / OWASP CRS compatible)

ndreno:
Records the WAF decision. Docs only, no code.

## The question

Whether Coraza's proxy-wasm filter could be dropped in opportunistically to give Barbacane a ModSecurity-compatible WAF.

It cannot. That binary targets Envoy's proxy-wasm ABI, which is unrelated to ours, and three runtime invariants would block any heavyweight plugin regardless. Those are now #138, and they turned out **not** to gate the route this ADR chooses.

## What it decides

A native Rust SecLang engine, compiled at `barbacane compile` time and sealed into the `.bca` artifact.

Five routes are compared and four rejected, each with the reason: implementing the proxy-wasm host ABI (about 50 host functions of permanent, externally-versioned surface for one plugin), cgo and C++ FFI (the ADR-0028 precedent, both link a foreign runtime into a data plane whose selling point is not having one), a Coraza sidecar over `host_http_call` (kept as the documented fallback, following `opa-authz`), and a TinyGo plugin on our own ABI (the retreat position).

The argument that settles it is failure mode rather than performance. A rule set is config of the same kind as a spec, so putting it through the same pipeline makes an unknown directive a **compile error** instead of a silently dropped rule. An engine that ignores what it cannot parse converts that rule into a bypass with no signal, and no runtime-loading WAF can offer the alternative. It also folds rule hashes into `artifact_hash` and makes spec-aware rule scoping possible, which is the part competitors structurally cannot copy: nobody else has a compile step or the contract.

## What was measured rather than assumed

The load-bearing assumption was that CRS needs PCRE. It does not. Against CRS v4.9.0, of 273 unique `@rx` patterns:

- 268 compile unmodified with the `regex` crate
- **zero** failures need lookaround or backreferences
- Go's RE2 accepts all 273, which places the gap in our parser's strictness rather than in the rule set
- a lexical repair pass reaches 273/273, differential-tested against RE2 over 297,129 inputs with no disagreements

The same sweep bounded the scope: CRS uses 18 operators, not the 35+ SecLang defines.

A survey of the Rust ecosystem is included, because "write it ourselves" is only defensible if nobody has. Nobody has: every Rust proxy advertising a WAF either binds libmodsecurity over C++ FFI, links Coraza through cgo, or ships a rule format that cannot consume CRS.

## The engine lives elsewhere, and the licence differs

[barbacane-dev/parapet](https://github.com/barbacane-dev/parapet), under `MIT OR Apache-2.0`. The ADR explains why.

Barbacane takes contributions under a DCO, which certifies the right to submit under the project licence but assigns no copyright and grants no relicensing right. Outside contributions to AGPL in-tree code therefore could not be offered under the commercial licence without per-contributor permission, and a SecLang engine is exactly the component that attracts outside contributions. A permissive engine dissolves that, and costs nothing: the engine is a commodity Coraza already gives away under Apache-2.0, while the compile-time integration is the differentiator and stays in this repository.

## Two corrections kept visible

Both are corrections to the ADR's own earlier drafts. They are left in the text rather than quietly fixed, because the reasoning is the useful part.

- **Compiled automata cannot be serialised into the artifact** as first claimed. Neither `regex` nor `aho-corasick` exposes a built program. The artifact carries the validated rule set and the data plane builds the automata once at startup, roughly 155 ms per process, which preserves the compile-time refusal, the hashing, and the elimination of per-request compilation.
- **libinjection was rejected, not adopted.** The first draft had it as the obvious next dependency. Audited against the C original, differential fuzzing keeps finding new divergence classes and one of them is a false positive, so the two operators that need it stay refused and fail loudly. Audit and reproduction steps are in the parapet repository.

## Status

Left as **Proposed**. Nothing in this repository depends on it yet; the next step is the compiler and data-plane integration, which is where it becomes real.

Refs #138

## Summary by CodeRabbit

- **Documentation**
  - Added an architectural decision record outlining potential Web Application Firewall (WAF) integration.
  - Documents a staged roadmap covering runtime improvements, an experimental Coraza-based integration, a native Rust engine, and future rule-scoping capabilities.
  - Includes feasibility findings, licensing considerations, competitive analysis, and open questions.


ndreno:
Thanks — five of the six findings were valid and are fixed in addbf97. Verified each against the code rather than taken on faith; notes on all six below.

**Valid, fixed:**

1. **GA gate contradicted itself.** Correct, and the sharpest catch. It asked for 100% of the CRS regression suite while the same ADR refused `@detectSQLi` and `@detectXSS`. Those four rules are 30 stages, so 100% was unreachable. I went the stricter way rather than defining a reduced suite: the two operators are now an explicit GA prerequisite, and GA requires the exclusion list to be **empty**, so the gate cannot be met by shrinking the target. The interim position is now stated instead of implied.

2. **`define_unknown_imports_as_trap` does not exist.** Correct. It is `define_unknown_imports_as_traps`, plural, taking `&Module` — verified in `wasmtime-47.0.3/src/runtime/linker.rs:261`. The corrected text also describes what it does, which is better than what the ADR originally asked for: unresolved imports get trapping stubs, so instantiation succeeds and the trap fires only if the guest calls one. #138 had the same error and is corrected too.

3. **P1 described the pooling, not the contract.** Correct, and it is the part that matters. Reuse needs exclusive checkout for the request, request-scoped state created fresh rather than carried over, and a defined reset before return to the pool. Without that one request can observe another's state, which is worse than the cost it fixes.

4. **`owasp-crs@4` is a floating alias.** Correct — two builds of the same spec would enforce different rules. The example now matches what the compiler actually implements (a path to a directory of SecLang files, resolved relative to the declaring spec, sealed and hashed into `artifact_hash`), and records that any future shorthand must resolve to an immutable pinned version.

5. **Measurements were not reproducible.** Fair. They now name CRS v4.9.0 by release tag, `regex` 1.12.3, Go 1.27.0, and the harness location, plus the fact that CI reruns the whole thing against the pinned version so the figures cannot drift from the code. Raw output stays in the harness rather than the ADR.

**Valid concern, narrowed:**

6. **`spec_scoped` shown enabled.** The concern is right: dropping WAF rules because the schema "already covers" that input class silently removes protection if the mapping is wrong, which is the exact failure this ADR exists to prevent. It now defaults to **off**, with the condition for enabling it written down (an audited rule-to-schema mapping, and keep the rule whenever coverage is uncertain).

   I did not commit to the full audited-allowlist-plus-coverage-tests machinery in the ADR. That is an implementation obligation for Stage 3, and specifying its shape before the feature is designed would be guessing. The default is off and the risk is recorded, which is what an ADR can honestly do here.

PR #137, opened 2026-08-10
chore(deps): bump wasmtime 43 -> 47 to fix RUSTSEC-2026-0222

ndreno:
## Why

The scheduled supply-chain audit ([run 31364240272](https://github.com/barbacane-dev/barbacane/actions/runs/31364240272)) failed on **RUSTSEC-2026-0222**, "Stores can mix up type indices between engines" ([GHSA-hgjw-h833-99q9](https://github.com/bytecodealliance/wasmtime/security/advisories/GHSA-hgjw-h833-99q9)), affecting `wasmtime` 43.0.2.

The 43.x line received no backport. The advisory's fixed ranges are `>=24.0.12 <25`, `>=36.0.13 <37`, `>=46.0.2 <47` and `>=47.0.3`, so the only forward move is the 47 line. Pinned to 47.0.3 in the lockfile.

## What changed

- `wasmtime` and `wasmtime-wasi` 43 -> 47 in the workspace dependencies, `Cargo.lock` updated
- Workspace MSRV 1.91 -> 1.94, since wasmtime 47 requires rustc 1.94. Propagated to the `msrv` CI job, the README badge, `CONTRIBUTING.md` and `docs/contributing/development.md`

No source changes were needed: the API surface we use (`Config`, `Engine`, `Module`, `Linker`, `Store`, `Caller`, `Memory`, `TypedFunc`, `ResourceLimiter`, `OptLevel`) is unchanged across 43 -> 47.

## Verification

Run locally before pushing:

- `cargo deny check advisories` -> **advisories ok**
- `cargo check --workspace --all-targets` -> clean
- `cargo clippy --workspace --lib --bins -- -D warnings` -> clean
- `cargo clippy --workspace --all-targets -- -D warnings -A clippy::unwrap_used -A clippy::expect_used -A clippy::panic` -> clean
- `cargo test -p barbacane-wasm` -> 195 passed, 0 failed, plus 1 doctest

## Out of scope

Two remaining cargo-deny warnings, neither of which fails the audit:

- `spin 0.9.8` is yanked, pulled in via `multer` <- `axum 0.8.8`. No upgrade path until axum bumps multer.
- Duplicate `winnow` 0.7.15 / 1.0.1 (`toml 0.8` on our side vs `toml 0.9` inside wasmtime). `bans.multiple-versions` is set to `warn`.

## Note for reviewers

The MSRV bump is the real cost here: building Barbacane now requires Rust 1.94.

## Summary by CodeRabbit

* **Chores**
[...]

PR #140, opened 2026-09-08
feat: WAF integration (ADR-0031 Stage 2)

ndreno:
Implements ADR-0031 Stage 2 (#139): a ModSecurity-compatible rule set, compiled at build time, sealed into the artifact, and evaluated as a native pipeline stage. Verified end to end against OWASP CRS v4.9.0.

Depends on #139 for the decision record. Engine lives in [barbacane-dev/parapet](https://github.com/barbacane-dev/parapet).

## What works

```
WAF enabled  rules=591  request_phase_rules=439  response_phase_rules=152
             blocking=true  paranoia_level=1  skipped_rules=[941100, 941101, 942100, 942101]

plain search   200      sqli union select  403      lfi traversal  403
numeric        200      xss script tag     403      rce $(id)      403
scanner UA     403      benign POST        200

WAF blocked: rule 949110 -> 403  inbound_score=18  "Inbound Anomaly Score Exceeded"
```

Blocking comes from CRS's own anomaly-score rule, not from individual rules denying, which is how a real deployment behaves.

## Design decisions worth reviewing

**A rule that cannot be enforced fails the build.** Shipping the rest of a rule set as though it were complete is how a rule becomes a bypass. That default makes stock CRS unbuildable today (four rules need `@detectSQLi`/`@detectXSS`), so `unsupported_rules: skip` is an explicit opt-in. Skipping is not silent: the compiler warns with the ids, the ids go in the manifest where `artifact_hash` covers them, the artifact carries only rules that will run, and the gateway logs them at WARN on every boot.

**The gateway refuses to start** when the artifact and the binary disagree about supported SecLang. Aligned with the fail-fast approach elsewhere; the alternative is serving traffic with rules missing and no signal.

**Both the rules and the policy are bound by `artifact_hash`.** The rule set goes in via `checksums`; the policy surface is hashed separately, because they are independently forgeable. Without that, a signed artifact could be flipped from blocking to detection-only or have its paranoia level lowered with the signature still valid. Tests assert each flip changes the hash.

**Paranoia level and thresholds are seeded into TX** before rules run. CRS gates entire rule files on the paranoia level and an unset variable reads as zero, so without seeding the rule set skips almost everything and *appears* to work.

**Rule id and message go to the log always, into the response only in dev mode.** In production they tell an attacker which rule to shape the next payload around.

## Known gaps, stated in the guide rather than buried

- **`@detectSQLi`/`@detectXSS` unimplemented**, so four CRS rules do not run. Regex-based SQLi and XSS rules still fire. The only pure-Rust libinjection port was [audited and rejected](https://github.com/barbacane-dev/parapet/blob/main/docs/audit-libinjectionrs.md); fixes contributed upstream in saarw/libinjectionrs#1.
- **Response-phase rules are not evaluated.** ~152 of 591 rules, mostly outbound data-leakage detection. The capability is implemented and tested but not wired: the response path has several exits and a streamed response reaches the client before a phase-4 rule could act. The gateway warns at startup rather than letting an operator assume they run.
- **~2 ms per request** with full CRS at PL1 (1.9 ms mean, 97% under 2.5 ms, measured through the gateway over 2,000 requests). One core sustains ~500 req/s of inspection, against ~1.2 µs for spec validation. Documented with that consequence, not as a footnote.

An isolated benchmark said 311-734 µs. That was the wrong number, not a wrong measurement: a tight loop keeps the rule set hot in cache and a server does not. The live figure is what the guide publishes.

## Notes for review

- `compile_inner` gained the spec directory: neither the spec tuple nor `ApiSpec::filename` carried it (`filename` is basename-only), so a relative `ruleset` path would have silently resolved against the process working directory. In-memory compilation passes `None` and refuses relative paths.
- `@pmFromFile` phrase lists are sealed alongside the rules and checksummed. Found by running the gateway: sealing only the rules left 23 CRS rules unresolvable and the gateway refused to start against the artifact it had just been handed. Artifact grows 56 KB → 188 KB, all phrase lists.
- New error code E1080, new warning W1080.
- `parapet` is a pinned git dependency per ADR-0031's sequencing; it moves to crates.io when its API settles.

Refs #138, #139

## Summary by CodeRabbit

* **New Features**
  * Added a ModSecurity-compatible web application firewall with OWASP CRS support.
  * WAF rules can block or monitor requests, inspect responses, and report matched rule IDs.
  * Added WAF rule and artifact integrity verification, including tampering and unknown-entry detection.
  * Compilation and startup now fail when WAF rules are missing, unreadable, or unenforceable.
  * Added metrics for WAF rule matches.

* **Documentation**
  * Added a Web Application Firewall guide covering configuration, limitations, tuning, metrics, and responses.


ndreno:
CI triage. Three failures, one mine.

**`sources FAILED` — mine, fixed.** cargo-deny denies unknown git origins and parapet is a git dependency. Allowlisted in `deny.toml` with the removal condition written next to it: the entry comes out when parapet moves to crates.io. `cargo deny check sources` now passes.

**`advisories FAILED` — pre-existing.** RUSTSEC-2026-0258 (hyper), 0268 and 0269 (wasmtime). Verified by running `cargo deny check advisories` against `main` in a clean worktree: it fails there too. The scheduled supply-chain audit has also been failing on main since 2026-08-31, independently of this branch. Needs its own chore PR, like #137 was; I have not touched it here because bundling a dependency bump into a feature PR makes both harder to review.

**`Clippy FAILED` — pre-existing, from a toolchain bump.** 47 `clippy::result_large_err` errors, of which **46 are in `barbacane-control`**, a crate this diff does not touch at all. The 47th is `main.rs:1962`, a handler signature that predates this branch (`git diff main...HEAD` shows I did not add it; my insertion only shifted its line number).

The lint got stricter in Rust 1.98, which CI picks up via `dtolnay/rust-toolchain@stable`. Locally on 1.96 the exact CI command passes:

```
$ cargo clippy --workspace --lib --bins -- -D warnings
Finished `dev` profile
```

Fixing it means boxing error variants across `barbacane-control`, which is a real change worth making but not in this PR: 46 unrelated edits would bury the WAF review. Worth its own issue, and it will block every PR until it lands.

Happy to open that issue, or to fold the fix in here if you would rather have it in one go.

PR #157, opened 2026-09-12
feat(waf): wire response-phase rules (CRS phases 3, 4, 5) into the pipeline

### LINKED ISSUE #145: WAF: wire response-phase rules (CRS phases 3 and 4) into the pipeline
## What is missing

`WafInspection::inspect_response` is implemented and unit-tested but never called. A CRS artifact carries around 152 rules in phases 3 to 5, mostly outbound data-leakage detection (stack traces, SQL error messages, source code, credentials in a response body), and they are sealed into the artifact, covered by `artifact_hash`, and never evaluated. The gateway warns about this at every boot.

Two consequences:

- Data-leakage detection, which is a large part of what CRS is for on the response side, does not happen.
- `x-barbacane-waf`'s `thresholds.outbound` is inert. The rule that compares against it (`959100` in CRS 4) is a phase 4 rule. It is still folded into `artifact_hash`, so changing it changes the artifact without changing behaviour.

## Feasibility: better than the doc comment claims

The comment on `inspect_response` says the response path "has several exits". That is misleading, and I checked both of the things that would actually have made this hard:

**There is a single convergence point.** `dispatch()` returns one `response` in the same scope where the `inspection` from `inspect_request` is still live, in `crates/barbacane/src/main.rs`. The other 15 return points in the handler are gateway-generated errors (validation failures, 404, 413, the WAF's own block) that need no response inspection. One hook site, not several.

**`Transaction` is `Send`.** parapet has no `Rc`, `RefCell` or `Cell` anywhere, and `set_request_body` copies rather than borrows, so the inspection can be held across the dispatch `await` without lifetime or auto-trait trouble.

## Proposed shape

The genuine constraint is narrower than "streaming makes this impossible":

**Phase 3 (response headers) works on every path, including streaming.** In the streaming dispatch path (ADR-0023) headers arrive through the channel *before* the body streams, so a phase-3 rule can still block. This covers CRS's response-header rules.

**Phase 4 (response body) works only on buffered responses.** A `Full` body reports an exact `size_hint().upper()` while a `StreamBody` reports `None`, which is a usable discriminator without threading a flag out of `dispatch()`. So:

- finite body under a size cap: collect, inspect, rebuild the response
- streaming or over the cap: skip phase 4, and **count the skip** rather than dropping it silently, the way `unsupported_rules: skip` records refused rules

Phase 5 is logging and cannot block, but CRS uses it for correlation, so it should still run.

WebSocket upgrades (101 then relay) have no meaningful response phases and should be excluded explicitly.

## Cost and open questions

- Adds a body collect to every buffered response under the cap. For buffered dispatch the body is already fully in memory as `Full<Bytes>`, so this is one pass plus a re-box, but it is on the hot path and wants measuring, not assuming. Request-phase inspection already costs about 2 ms with full CRS; the response side should be measured separately rather than assumed symmetrical.
- What the size cap should be, and whether it is configurable.
- Whether a phase 4 skip on a streamed response should be visible to the operator as more than a counter, given that the guide currently promises outbound inspection once this lands.

## Testing

`crates/barbacane-test/tests/waf.rs` and `tests/fixtures/waf-rules` already exist and the fixture rule set is in CRS 4 idiom, so response-phase rules can be added to the fixture and asserted end to end: a leaking buffered response blocked, a clean one untouched, a streamed response inspected for headers and counted-but-not-blocked for its body.

Four of the eight unit tests in `crates/barbacane/src/waf.rs` already cover `inspect_response`, including the streamed no-body case, so the engine side is tested ahead of the wiring.

## Related

- ADR-0031 (WAF integration), #140, #143
- `docs/guide/waf.md` documents the limitation and the inert threshold

ndreno:
Closes #145.

## What

`WafInspection::inspect_response` was implemented and unit-tested but never called, so the ~152 phase 3-5 rules a CRS artifact carries (mostly outbound data-leakage detection: stack traces, SQL errors, source code, credentials in a response body) were sealed into the artifact and never evaluated. `thresholds.outbound` was inert for the same reason.

This wires response inspection into the pipeline at the single convergence point after `dispatch()`, continuing the same transaction the request phase started (so outbound blocking rules read the scores the inbound rules accumulated).

## Behaviour

- **Phase 3 (response headers)** runs on every path, including streaming.
- **Phase 4 (response body)** runs on a buffered body at or under `max_response_body` (new manifest field, default 1 MiB, configurable). A streamed or oversized body has its headers inspected and its body skipped, counted in the new `barbacane_waf_response_body_skipped_total` metric rather than dropped silently.
- **Phase 5 (logging)** always runs.
- A **WebSocket upgrade (101)** has no response phases and is excluded.

`max_response_body` is covered by `artifact_hash`. The boot-time warning that response rules were present-but-not-evaluated is removed.

## Tests

Five new integration tests in `crates/barbacane-test/tests/waf.rs`:
- a leaking response body is blocked (phase 4)
- a leaking response header is blocked (phase 3)
- a clean response passes through
- an oversized body is not inspected and the skip is counted
- response headers are still inspected when the body is skipped

New fixture rules (`030-response-rules.conf`, `040-response-blocking.conf`) mirror CRS `959100` on the outbound score, and a `waf-response-cap.yaml` spec with a tiny cap exercises the skip path. The existing `inspect_response` unit tests already cover the engine side.

## Docs

`docs/guide/waf.md` gains a "Response-phase inspection" section and the skip metric; the "not evaluated" limitation and the "outbound is inert" note are removed. ADR-0031's open question on phase-4 vs streaming is marked resolved.

## Summary by CodeRabbit

- **New Features**
  - Added WAF inspection for response headers and buffered response bodies.
  - Responses exceeding the configured body-size limit or using streaming are skipped for body inspection, while headers remain inspected.
  - Added configurable response-body inspection limit, defaulting to 1 MiB.
  - Added metrics tracking skipped response-body inspections.

- **Documentation**
  - Updated WAF documentation with response-phase behavior, configuration, and metrics.


ndreno:
Thanks @coderabbitai. Addressed in 3c77e8c:

**1. `max_response_body` checked conversion** — done. `WafStage::from_artifact` now uses `usize::try_from(...).unwrap_or(usize::MAX)` instead of `as usize`, so a cap larger than the address space saturates to "no effective limit" rather than wrapping to a small value that would silently under-inspect. Phase-4 inspection never receives a wrapped or zero cap.

**2. Phase 5 on a blocked response** — corrected, but not the way the finding suggests. The premise is ModSecurity's convention that phase 5 (logging) always runs even after an interception. This engine (parapet) deliberately does not implement that: `Transaction::run_phase` returns early when `verdict != Allow`, so a disruptive verdict stops *all* later phase processing, phase 5 included. My wrapper cannot make phase 5 run after a block by reordering, and claiming it does would be false. `inspect_response` now runs the phases in order and lets the engine short-circuit, with the doc comment stating the actual model. Added regression tests: `phase_five_runs_on_a_non_blocked_response` and `a_phase_three_block_stops_before_later_phases`. (parapet's divergence from ModSecurity here only has observable effect once audit logging / persistent collections are wired, neither of which exists yet; tracking separately.)

**3. Streaming-dispatcher response test** — the skip path is already covered. A true stream (`size_hint().upper() == None`) and an oversized body (`Some(n > cap)`) both take the same `else` arm in `inspect_response_waf`: skip metric recorded, phase-3 headers still inspected, no phase-4 body match. `an_oversized_response_body_is_not_inspected_and_the_skip_is_counted` and `response_headers_are_inspected_even_when_the_body_is_skipped` assert exactly that. Standing up a full upstream + wiremock streaming harness (no existing WAF test uses an upstream) to vary only the size-hint discriminator would be disproportionate for the same branch.

PR #166, opened 2026-09-14
docs(seo): add sitemap, robots.txt, and Open Graph tags

ndreno:
The docs SEO pass (#4 from the Search Console analysis). The mdBook docs site had **no sitemap.xml and no robots.txt**, so engines discovered pages only by crawling links. The main website has a sitemap via Astro; the docs had neither.

## Changes
- **`docs.yml`**: after the mdBook build, generate `book/sitemap.xml` (every rendered page, excluding the print bundle and 404) and `book/robots.txt` pointing at it. Both ship to GitHub Pages with the site.
- **`book.toml`**: point the theme dir at `docs/theme` so an additive `head.hbs` is picked up.
- **`docs/theme/head.hbs`**: Open Graph + Twitter Card tags (title, description, image) so shared and crawled doc links look right. mdBook merges this into the default theme, it does not replace it.

## What this does and does not do
- Sitemap + robots is the real lever here: it helps Google discover and recrawl all doc pages, which is the main constraint on a link-crawled static site.
- mdBook does **not** support per-page meta descriptions natively (the default theme uses the book-level description everywhere), so this pass does not add unique per-page descriptions. Google generates doc snippets from page content anyway, and the doc intros are already clear, so I did not rewrite content for SEO.
- Page titles come from the SUMMARY.md link text and are already descriptive, so I left them.

## Validation
Built locally with mdBook 0.5.4 (matches CI): 31 URLs, well-formed sitemap (correct `sitemaps.org` namespace), robots.txt references it, OG tags present in the built HTML. The docs workflow only runs on push to `main`, so it could not be exercised on this PR; the local build stands in.

## Follow-up (manual)
Submit `https://docs.barbacane.dev/sitemap.xml` in Search Console once merged, so indexing picks it up faster.

## Summary by CodeRabbit

- **Documentation**
  - Added automatic sitemap and robots.txt generation for the documentation site.
  - Sitemap entries include built documentation pages and their last-updated date.
  - Added SEO and social-sharing metadata, including Open Graph and Twitter card information.
  - Documentation pages now use the configured custom theme for consistent page metadata and presentation.


ndreno:
Good catch @coderabbitai. Fixed: `lastmod` now comes from each source page's last commit date (`git log -1 --format=%cs`), with `fetch-depth: 0` on the checkout so the history is there, falling back to the build date only when a page has no matching source file. Validated locally: pages carry distinct real dates (e.g. dispatchers 2026-07-21, waf 2026-09-12) instead of a uniform today's date.

Issue #164, opened 2026-09-12
CI/merge hygiene: 'Workflows' check hangs pending; stacked-PR base deletion closes dependents


ndreno:
Two related pain points surfaced while merging the 0.10.0 work (#157, #159/#160, #161, #162). Both forced manual intervention.

## 1. The `Workflows` (zizmor) check reports pending after its run succeeds

The `workflows` job in `ci.yml` (name: `Workflows`, runs zizmor) reliably shows as `pending` in `gh pr checks` and keeps the PR `mergeStateStatus: BLOCKED` **even after its workflow run has concluded `success`** (the job has `conclusion=success` and a `completedAt`, but the check status never reconciles). Observed on every PR in the 0.10.0 batch, each requiring an `--admin` merge.

Effect: no PR can auto-merge on green; the required check never reports green to branch protection.

Directions to investigate:
- The job runs on `push` with no `needs`/paths filter, so it may register a check context that branch protection tracks differently from the PR head SHA (duplicate/hung context).
- Confirm whether the required-check name in branch protection matches exactly one context.
- Consider gating it behind the `changes` filter or giving it a stable check name, or moving the workflow-audit into the `fmt`/`clippy` fast job.

## 2. Deleting a base branch on merge closes stacked PRs instead of retargeting

Merging #157 with `--delete-branch` while #159 was stacked on `feat/waf-response-phase` **closed** #159 rather than retargeting it to `main`; it had to be reopened as a fresh PR (#160) after rebasing off the now-merged base. GitHub retargets dependent PRs only in some cases; a deleted base branch generally closes them.

Mitigations:
- Open stacked PRs against `main` from the start, or
- Merge the base PR **without** `--delete-branch` until the stacked PR is retargeted, then delete, or
- Document the sequence in `CONTRIBUTING`/release notes.

Neither is a correctness bug in the gateway; both are CI/release-process hygiene that cost time this cycle.

ndreno:
Root-caused. Two separate things were conflated under "the Workflows check hangs."

**1. The spurious "Workflows" check (fixed in #167).** `ci.yml` triggers on `issue_comment` (for `/bench`), and the `workflows` (zizmor) job had no event guard, so every comment on a PR fired a CI run whose only active job was zizmor. Those comment-triggered runs are what produced confusing "Workflows" check runs. #167 adds `if: github.event_name != 'issue_comment'` to the job; now a comment runs nothing unless it is an authorized `/bench`.

[...]

PR #169, opened 2026-09-14
ci: single "CI" gate job for branch protection

ndreno:
Follow-up to the #164 diagnosis. Replaces the brittle per-shard required-check list with one stable context.

## Why
Branch protection matches required checks by **exact context name**. When #154 sharded the integration job, the contexts became `Integration Tests (shard 0..3)` and the required `Integration Tests` matched nothing, blocking every PR. The same brittleness applies to `Build`/`Unit Tests` on docs-only PRs (they skip and rely on "skipped counts as pass").

## What
A single `ci-success` job (context name **`CI`**) that:
- `needs` every gating job (fmt, clippy, workflows, msrv, audit, build, unit-tests, control-plane-tests, plugin-tests, integration-tests, coverage, security-suite, ui-tests, ui-e2e),
- runs with `if: always()` so it evaluates even when needed jobs were skipped,
- fails only if a needed job's result is `failure` or `cancelled`; `success` and `skipped` pass,
- does not run on `issue_comment`.

Excludes `changes` (the enabler) and `benchmarks` (opt-in, informational).

## After merge
I'll switch branch protection to require the single `CI` context and drop the four `Integration Tests (shard N)` entries. Then a shard-count change or a job rename never breaks the merge gate again, and docs-only PRs pass cleanly.

## Summary by CodeRabbit

* **Chores**
  * Added a single consolidated CI status check that reports the outcome of all required validation jobs.
  * The check now clearly fails when any required job fails or is cancelled, simplifying build and pull request status tracking.


ndreno:
Good catch @coderabbitai, that's a real hole. Added `changes` to the gate's `needs` (a37a9f6). Now a change-detection failure surfaces as a `failure` result and fails the `CI` gate, instead of skipping the code jobs and being read as success.

PR #189, opened 2026-09-16
refactor(plugins): read and write the request context through the SDK

ndreno:
## Problem

Eleven plugins each carried their own copy of the `host_context_get` / `host_context_set` bindings: a `#[cfg(target_arch = "wasm32")]` extern block with its `unsafe` pointer arithmetic, a native stub, and a `CONTEXT` thread_local in the test mock. The same contract with the host, restated eleven times, with eleven places for it to drift.

This was the largest duplication cluster in the repo, found while measuring it: roughly 620 lines of near-identical binding code.

## Change

`barbacane_plugin_sdk::context` already provides `get`, `set` and `clear`, with the same wasm bindings and a thread-local native store. The auth plugins and `rate-limit` moved to it when it was introduced; these eleven now follow:

`ai-cost-tracker`, `ai-prompt-guard`, `ai-response-guard`, `ai-token-limit`, `ai-proxy`, `cel`, `correlation-id`, `cors`, `http-log`, `observability`, `request-transformer`.

Each `mock_host` keeps its other thread_locals (logs, counters, clock, uuid, http calls) and its `reset()` clears the context through the SDK, so existing test code is unchanged. Every other host binding was left alone.

**462 lines removed.** Duplication over `crates` + `plugins`, measured with jscpd at the same settings before and after: **3.14% (90 clones) to 2.62% (75 clones)**.

## The one deviation worth review

`cel` and `ai-proxy` had a `get_context() -> BTreeMap` test helper that dumped the whole map. The SDK exposes no map dump, so it became a per-key `context_value(key) -> Option<String>` and those assertions read one key each: `ctx.get(k).map(|s| s.as_str()) == Some(v)` became `context_value(k).as_deref() == Some(v)`, and `!ctx.contains_key(k)` became `context_value(k).is_none()`. Assertion counts are identical (cel 49, ai-proxy 136) and no expected value changed.

## Not changed

- **`plugin.toml`**: the plugins bind the same host functions, just through the SDK, so capabilities are identical.
- **Behaviour**: pure deduplication.

The plugin `Cargo.lock` files pick up the `barbacane-plugin-sdk` path dependency at its current 0.10.0. They were pinned at 0.7.0, which every build was already having to resolve past.

## Verification

Per plugin, from its own directory: `cargo fmt`, `cargo clippy --all-targets`, `cargo test`, and `cargo check --target wasm32-unknown-unknown`. The wasm check is the one that matters, since that is the target whose bindings were removed.

Test counts are unchanged, 419 in total:

| Plugin | Tests | Plugin | Tests |
|---|---|---|---|
| ai-cost-tracker | 13 | cel | 56 |
| ai-prompt-guard | 29 | correlation-id | 10 |
| ai-response-guard | 24 | cors | 35 |
| ai-token-limit | 31 | http-log | 12 |
| ai-proxy | 134 | observability | 17 |
| request-transformer | 58 | | |

Two pre-existing clippy warnings in `ai-proxy` (`unnecessary_get_then_check`) and two in `http-log` (`thread_local can be made const`) are on untouched lines and remain. CI runs `cargo test` per plugin, not fmt or clippy.

## Summary by CodeRabbit

- **Improvements**
  - Standardized request-context handling across AI protection, proxy, token-limit, transformation, CORS, logging, observability, correlation ID, and policy features.
  - Context values used for routing, profile selection, metadata, token limits, request transformation, and response processing now behave consistently across supported environments.
  - Existing request, response, authentication, logging, metrics, and translation behavior remains unchanged.


ndreno:
@coderabbitai full review

ndreno:
@coderabbitai full review

ndreno:
@coderabbitai full review

ndreno:
@coderabbitai full review

PR #190, opened 2026-09-16
fix(compiler): resolve $ref parameters instead of dropping them

ndreno:
Step 0 of ADR-0033. Also a standalone correctness fix.

## Problem

`parse_parameters` required an `in` key on every entry of a `parameters` array:

```rust
let Some(location) = param_obj.get("in").and_then(|v| v.as_str()) else {
    continue;   // silently skipped
};
```

An entry of the form `- $ref: "#/components/parameters/TenantHeader"` carries `in` and `name` on the *target*, not on the entry, so it hit that `continue`. Nothing resolved `#/components/parameters/`, and nothing warned. The parameter was absent from the compiled operation, so a spec that factors its shared parameters into `components` had **no validation at all** on them: required-ness and schema both ignored.

This affects all three parsing sites (path item, operation, and OpenAPI 3.2 `additionalOperations`), since they funnel through this one function.

## Why it blocks ADR-0033

The request header allowlist is derived from declared parameters. A `$ref`'d header parameter that the compiler drops becomes a header the **data plane drops at runtime**: a silent validation gap turns into a silent connectivity failure. This had to be fixed before the allowlist is built on top.

## Change

Parameter entries are resolved through the reference chain before being read, with cycle detection, mirroring `resolve_schema_refs`. A reference that does not resolve is now an error (`ParseError::UnresolvedRef`) rather than a silent omission, matching how schema references already behave.

Note this is a behaviour change for a malformed spec: a dangling parameter `$ref` previously compiled and now fails. That is the intended direction, and it cannot break a spec whose references are valid.

## Tests

Four new tests in `spec_parser::parser::tests`, all verified to run (148 pass, up from 144):

- `resolve_ref_to_components_parameters`: a `$ref` at both path-item and operation level resolves, covering a header and a cookie parameter, with `required` and schema preserved.
- `resolve_chained_ref_to_components_parameters`: a reference to a reference resolves to the end of the chain.
- `unresolvable_parameter_ref_is_an_error`: a dangling `$ref` is no longer ignored.
- `circular_parameter_ref_is_an_error`: a self-referential parameter errors instead of hanging.

## ADR-0033 amendments

The ADR is amended in the same commit so the decision matches the code it will be built on. Four corrections, all found by surveying the existing code:

1. **`x-real-ip` joins the baseline.** `ip-restriction` and the `client_ip` partitions of `rate-limit` and `ai-token-limit` read it; the original baseline listed only `x-forwarded-*` and `forwarded`, so those plugins would have broken.
2. **The security scheme model is called out as new work.** Neither `components.securitySchemes` nor any `security` block is parsed today, and no Rust type models the scheme kinds, so set 3 of the allowlist is greenfield rather than a lookup. The section also records the resolution rule (operation overrides root, `security: []` is anonymous) and notes `E1040` is already taken by `UndeclaredPlugin`.
3. **The artifact section is corrected.** `ARTIFACT_VERSION` is never compared at load today, so bumping it alone changes nothing: rejection of an old artifact happens only incidentally through the hash. The decision now includes an explicit version check so a format mismatch reports its real cause instead of "integrity check failed".
4. **The WAF header list stays unfiltered**, and URL-sourced plugins are noted as contributing no configured header names, the same gap that already limits secret-field detection.

## Summary by CodeRabbit

- **New Features**
  - OpenAPI parameter definitions can now reference reusable parameters, including chained references.
  - Added support for the OpenAPI 3.2 `querystring` parameter location.
  - Proxy handling now recognizes the `x-real-ip` header for client IP–based restrictions and rate limits.

- **Bug Fixes**
  - Improved validation and error reporting for unresolved or circular parameter references.
  - Added data-plane validation to detect incompatible artifact versions during loading.


ndreno:
@coderabbitai full review

ndreno:
Fixed in bad8ea1. Verified the finding against the code before acting: the merge was `path_params.clone()` then `extend()`, and `OperationValidator::new` (validator.rs:394-414) pushes every entry into its per-location list, so a duplicate (name, in) really was validated twice and a stale inherited schema could reject a value the operation accepts.

`merge_parameters` now drops any inherited parameter the operation redefines, applied at both call sites: the path-item walk and the OpenAPI 3.2 `additionalOperations` walk, which carried the same inline merge you flagged.

One addition beyond the report: header names are compared case-insensitively, so `X-Tenant-Id` on the path item and `x-tenant-id` on the operation are recognised as the same parameter rather than both surviving. Other locations stay case-sensitive.

Test `operation_parameter_overrides_the_path_item_one` covers a query override (asserting the operation's `required` and schema win), a case-differing header override, and an inherited path parameter the operation does not redefine, with a total count so a duplicate cannot slip through.

PR #203, opened 2026-09-17
chore(release): 0.11.0

ndreno:
Closes the Unreleased section and bumps the workspace to 0.11.0. Three files: the changelog heading, `Cargo.toml` and `Cargo.lock`.

Rebuilt on top of main rather than rebased. The branch predated #199, #202, #205 and #206, so a rebase would have reverted eighty-two files.

## Headline

A request carries only the headers its operation admits, so the document describes what an upstream receives and not only what a client may send. Specs that could not be compiled at all now can, including any with a recursive schema.

## What a user has to do

Three things, each with its own upgrade note in the changelog:

- **Recompile every artifact.** `routes.json` changes shape and its checksum is bound into `artifact_hash`. The format version is 6 and the data plane compares it, so a stale artifact is refused by version with a message naming the recompile rather than failing the integrity check.
- **Declare the headers your upstream needs.** A header the document does not describe no longer reaches it. `barbacane_request_headers_dropped_total` counts what was dropped, and `serve --dev` logs each one by name, so a first run against an existing spec shows exactly what to declare.
- **Declare the security scheme where a middleware authenticates.** An operation attaching `basic-auth`, `jwt-auth`, `apikey-auth`, `oauth2-auth`, `oidc-auth` or `ldap-auth` must name a scheme defined under `components.securitySchemes`. Compiling one that does not is `E1057`, naming the operation, the plugin and the fix.

## After merge

Tagging `v0.11.0` publishes binaries, containers and crates.io, and is left to a human.

The release workflow builds every plugin from source with no cache, which it must: a plugin's `plugin.toml` and `config-schema.json` are embedded into its `.wasm` at build time, so prebuilt binaries would carry the previous metadata silently.

## Summary by CodeRabbit

- **Breaking Changes**
  - Requests now include only headers explicitly permitted by their operation.
  - Headers not described by the document are no longer forwarded upstream.
  - Authentication-middleware operations must declare the security scheme carrying credentials.
  - All `.bca` artifacts created before version 0.11 must be recompiled due to the updated artifact format.

- **Release**
  - Version 0.11.0 is now available.


ndreno:
@coderabbitai full review

PR #205, opened 2026-09-17
feat(compiler): collect the header names a plugin's configuration reads

ndreno:
ADR-0033 set 4, the last part of the allowlist that is **derived rather than enforced**. Enforcement is the next change.

## Why this comes before enforcement

Step 2 derived sets 2 and 3, which come from the spec: declared parameters and security schemes. Set 4 is the headers a plugin's own **configuration** tells it to read, and nothing in the spec mentions them. Enforcing without this would silently break:

| Plugin | Configured header |
|---|---|
| `apikey-auth` | `header_name`, the key header |
| `correlation-id` | `header_name` |
| `rate-limit`, `ai-token-limit` | the header a `header:<name>` partition selects |
| `cache` | the `vary` list |
| `kafka`, `nats` | `headers_from_request`, and kafka's `key` template |
| `cors` | `allowed_headers` |
| `request-transformer` | the headers it removes or renames |

## The annotation

A plugin marks such a field in its own `config-schema.json`, the way `writeOnly` already marks a secret. **Three** formats, because the shapes genuinely differ:

- **`header-name`**: the value is a header name, or a list of them.
- **`header-ref`**: the value may *reference* one among other things, as a partition key does with `header:<name>` or a message key with `$request.header.<name>`.
- **`header-name-map`**: the object's **keys** are the names, as in a rename table.

The second format exists because a test caught a design flaw: one format cannot mean both "this string **is** a header name" (`header_name: X-API-Key`) and "this string **may reference** one" (`partition_key: client_ip`, which selects something that is not a header at all). Collapsing them would have admitted `client_ip` as a header name.

The third matters for a reason worth stating: in a rename table the **keys** are read and the **values** are written. Admitting the values would let a client supply a header the plugin is supposed to produce.

The configuration is walked **alongside the schema**, so a marked field is read where it actually sits rather than wherever its name happens to appear. That matters for `request-transformer`, whose marked fields are nested under `headers`.

## Plumbing

The schema now reaches the compiler on `PluginBundle`. A URL-sourced plugin still has none, exactly as it has no secret fields today, so headers it is told to read must be declared in the spec like any other. The control plane is in the same position, since the registry does not persist `config-schema.json`. Both are noted in the code rather than left to be discovered.

## Tests

176 unit tests and 7 corpus checks. Four cover the collector directly, one per shape, including that an **unmarked** field is not read and that a rename's new name is not admitted. One is end to end: it compiles `rate-limit.yaml`, reads `routes.json` back out of the artifact, and asserts `x-client-id` reached the allowlist because the fixture partitions on `header:x-client-id`.

The workspace check caught the third `PluginBundle` construction site, in the control plane, which a crate-scoped test would have missed.

The vacuum ruleset is unchanged and its 16 tests pass: the generator drops `format`, so the annotations do not affect it. Verified rather than assumed.

## Summary by CodeRabbit

- **New Features**
  - Plugin configurations can now identify request headers they use, allowing those headers to be included automatically in compiled route permissions.
  - Header references and header-name settings are recognized across authentication, caching, CORS, messaging, rate limiting, and request transformation plugins.

- **Bug Fixes**
  - Prevented plugin-configured request headers from being unintentionally blocked during request processing.

- **Validation**
  - Expanded configuration schemas with header-specific format annotations for clearer and more consistent validation.


ndreno:
Addressed in fd5fb7e (force-pushed).

**1. Reject reserved identity headers from plugin configuration** — fixed. You were right that the rule was applied inconsistently: `reserved_checked` guarded declared parameters and security schemes, but a configured name went straight into the set. `push_header_name` now delegates to `reserved_checked` rather than repeating the rule, and `extract_header_names`, `extract_referenced_header_names` and `collect_configured_header_names` return `Result` so a compile fails with E1056.

Added `configured_headers::reserved_identity_headers_are_refused`, covering all four marked formats plus the selector and template syntaxes. I verified it is non-vacuous by disabling the rule in `reserved_checked`: both reserved-name tests fail, and pass again once restored.

Worth noting for severity: the data plane already drops `x-auth-*` unconditionally regardless of the allowlist, so this was not exploitable at runtime. The defect was that the derived list disagreed with the rule the compiler enforces elsewhere.

**2. Fail on fixture compilation errors** — fixed, and the broad skip was hiding more than style. The test passed `manifest_path` where `compile_with_manifest` expects `manifest_base_path`, a directory. Plugin paths resolved as `tests/fixtures/barbacane.yaml/../../plugins/<name>/<name>.wasm`, which traverses through a regular file, so compilation returned `ENOTDIR` on every machine:

```
failed to read plugin 'basic-auth' from .../barbacane.yaml/../../plugins/basic-auth/basic-auth.wasm: Not a directory (os error 20)
```

The `Err` arm caught that and returned early, and `cargo test` hides stdout for passing tests, so the test reported green while asserting nothing. It now checks the fixture's two plugins are built and otherwise propagates the error. With the base path corrected it passes for real.

**3. Make `header_name` conditional on `key_location`** — declining, because it would reintroduce the bug this feature exists to prevent. `collect_configured_header_names` walks `properties`, `items` and `additionalProperties`. It does not evaluate `if`/`then`/`allOf`/`oneOf`, so moving `header_name` under a conditional branch makes it invisible to the walk and the header stops being admitted in `header` mode as well, which is the common case.

The allowlist is deliberately a superset. Over-admitting in `query` mode forwards one header the operator named in their own configuration, and the plugin ignores it. Under-admitting drops a header the plugin needs and breaks the request. Evaluating conditionals in the walk is a reasonable future change; gating on a keyword the walk cannot read is not.

PR #211, opened 2026-09-18
fix: stop dropping tokio runtimes inside the async context

### LINKED ISSUE #208: Dropping a runtime inside the async context panics: URL plugin compiles, and every graceful shutdown
## Symptom

`barbacane compile` panics whenever it has to download a plugin, when built with the `dev` profile. Deterministic, three of three runs with the cache bypassed.

```
thread 'main' panicked at tokio-1.51.1/src/runtime/blocking/shutdown.rs:51:21:
Cannot drop a runtime in a context where blocking is not allowed.
This happens when a runtime is dropped from within an asynchronous context.
```

Backtrace:

```
barbacane_compiler::download::build_client        download.rs:32
barbacane_compiler::download::download_plugin     download.rs:65
barbacane_compiler::manifest::resolve_url_plugin  manifest.rs:304
barbacane_compiler::manifest::resolve_plugin      manifest.rs:200
ProjectManifest::resolve_used_plugins             manifest.rs:566
```

## Reproduction

```bash
mkdir repro && cd repro
cat > barbacane.yaml <<'YAML'
plugins:
  mock:
    url: https://github.com/barbacane-dev/barbacane/releases/download/v0.11.0/mock.wasm
    sha256: 24a889e2fe3c1d1bb30520c9d60bd8a1df82252fc5ce952769f761bb1a0b7b66
YAML
cat > api.yaml <<'YAML'
openapi: "3.0.3"
info: { title: repro, version: "1.0.0" }
paths:
  /ping:
    get:
      operationId: ping
      x-barbacane-dispatch:
        name: mock
        config: { status: 200, body: "pong" }
      responses: { "200": { description: ok } }
YAML

cargo run -p barbacane -- compile -s api.yaml -m barbacane.yaml -o out.bca --no-cache   # panics
cargo run --release -p barbacane -- compile -s api.yaml -m barbacane.yaml -o out.bca --no-cache   # succeeds
```

`--no-cache` matters: with a warm plugin cache no download happens and the panic does not occur, which is why this is easy to miss.

## Cause

`main` is `#[tokio::main]`, so `compile` runs inside the runtime. `build_client` (`crates/barbacane-compiler/src/download.rs:27`) constructs a `reqwest::blocking::Client`, which owns an inner runtime. Dropping that inner runtime inside an async context is what tokio refuses.

## Why release builds appear unaffected — unconfirmed

Verified the difference is the build profile, not the environment:

| build | cold downloads | panics |
|---|---|---|
| dev, local | 3 | 3 |
| release, local, same tree | 3 | 0 |
| release, published `v0.11.0` binary | 5 | 0 |

Source and `Cargo.lock` are byte-identical (tokio 1.51.1 in both), same machine.

**The mechanism is not established.** Note the `panic!` in tokio is *not* behind `cfg(debug_assertions)` — it is gated only on whether the calling thread may block. So release immunity is probably incidental rather than guaranteed, and a plausible explanation is timing: if the client's background runtime has already gone idle by the time it is dropped, the shutdown completes without needing to block. If that is right, a release build could hit this under a slow network or heavy load. I could not produce it in release, so this stays a hypothesis.

Either way the code is doing something tokio considers invalid, so it is worth fixing regardless of which profiles happen to survive it.

## Not a regression

The same shape exists in `v0.10.0`: `build_client` returns a `reqwest::blocking::Client` and `main` is `#[tokio::main]`. Pre-existing, not introduced by 0.11.0.

## Suggested fix

Stop mixing a blocking client into the async runtime. Either use `reqwest`'s async client for plugin downloads, or move the whole resolution onto a plain `std::thread` so the inner runtime is created and dropped outside the async context. `spawn_blocking` is not sufficient, since the problem is the drop, not the request.

A regression test should force a real download, since a warm cache hides it. `--no-cache` is the lever.

ndreno:
Closes #208. Three sites, all in the released binary, all the same panic.

## Shutdown, the one that mattered

`barbacane serve` panicked on **every** graceful shutdown and exited **101**:

```
barbacane: all connections drained, shutting down
thread 'main' panicked at tokio .../blocking/shutdown.rs:51:21:
Cannot drop a runtime in a context where blocking is not allowed.
```

The drain was correct; only the exit code lied about it. That is how a supervisor decides a process crashed, so under Kubernetes, compose or systemd every clean stop produced restart backoff, crash-loop counters and alerts.

**Cause.** `Gateway::load` builds a NATS publisher, a Kafka publisher and an LDAP client **whatever the artifact contains**, each carrying its own tokio runtime for the synchronous calls WASM host functions make. A mock-only artifact still carried three. Letting the gateway fall out of scope at the end of `run_serve` dropped all three inside the runtime.

Worth noting separately: three unused runtimes per process is also waste, and making them lazy is a reasonable follow-up. This change fixes the panic without changing when they are built.

## Startup failure

The same drop on every early return after the artifact loads: an invalid address, a taken port, an unreadable TLS config, a bad admin bind. The real cause was printed and then buried under a panic, and a configuration error became indistinguishable from a crash by exit code.

## Compiling a URL-sourced plugin

`reqwest::blocking` refuses to run inside a tokio runtime, both on construction and on every request, and `compile` runs inside one. A manifest with a remote plugin could not be compiled at all from a dev-profile build. The download runs on a scoped thread, which can still borrow the URL.

I found this one twice over: my first attempt kept the client in a `OnceLock` to avoid dropping it, and the panic simply moved to construction; the second moved construction off-runtime, and it moved to `execute_request`. Only taking the whole call off the runtime works.

## Verified

| site | before | after |
|---|---|---|
| SIGTERM on a healthy gateway | exit 101, panic | exit 0, clean drain |
| taken port | exit 101, panic hiding the cause | exit 1, cause reported |
| URL plugin compile | panic, both profiles | compiles, both profiles |

## The test that should have existed

`process_lifecycle` spawns the binary, waits for it to serve, sends SIGTERM and asserts a clean exit; a second test holds a port and asserts the cause is reported without a panic.

Nothing did this before. Every other test drives the gateway in-process, which is exactly why this shipped. The test fails against the old code with the panic in stderr, so it is not vacuous.

## Summary by CodeRabbit

* **Bug Fixes**
  * `barbacane serve` now shuts down cleanly with exit code 0 when receiving `SIGTERM`, without panicking.
  * Startup failures, such as unavailable listening ports or invalid configuration, now report the underlying error without panicking.
  * `compile` now successfully resolves plugins provided through URLs.


ndreno:
Two of the three fixed in 946c759; declining the first, with evidence.

**Join gateway-owning tasks before final teardown (Major)** — correct, and fixed at a better level than suggested. `SharedGateway` is an `Arc` that the MCP eviction and hot-reload tasks clone, neither is joined, so either could hold the last reference and drop it on a worker after the local one is gone. My change worked only because of ordering, which is worse than not working: the panic would have returned rarely rather than reliably.

Rather than retain and join the handles, I fixed it where the runtimes live. `NatsPublisher`, `KafkaPublisher` and `LdapClient` now hand their runtime to `shutdown_background()` in `Drop` instead of waiting for workers to stop. The last reference is then safe to fall anywhere, including a worker thread, and it holds for any future holder rather than only the two tasks that exist today.

Tests drop one inside `block_on` and inside a spawned task. Both fail with the original panic if the plain drop is restored.

**Add a regression test for a real HTTPS plugin download** — correct, that gap was real. The lifecycle test builds its fixture from a local plugin path, so nothing reached `download_plugin`.

Added two tests that call it from inside a runtime and from a spawned task, against a `.invalid` host so they fail fast and need no network. They assert an error is returned rather than the process dying, which is exactly what regressed. Both fail without the scoped thread, one with `Cannot drop a runtime...` and the other with `JoinError::Panic`.

I chose an unreachable host over a real download deliberately: the bug was reqwest refusing to run inside a runtime at all, which reproduces before any byte moves, and a test that needs the network would be flaky for no extra signal.

**Keep the profile directory in `gateway_binary`** — declining; the count is right as written.

`current_exe()` for an integration test is the executable itself, not a directory:

```
target/debug/deps/process_lifecycle-3352eed035cf2162
```

The first `pop()` removes the file name, giving `target/debug/deps`; the second gives `target/debug`; the join gives `target/debug/barbacane`, which exists.

Verified two ways: the run prints no `skipping` line, and reverting the shutdown fix makes the test fail with the panic in stderr. A test that could not find the binary would have returned early and passed.

ndreno:
@coderabbitai three findings, addressed in 57dd703.

**Join gateway-owning tasks before final teardown** (major). The race is real: a detached eviction or hot-reload task holding a `SharedGateway` clone can drop the last reference on a worker thread. It is not a panic, because each runtime-owning client hands its runtime to `shutdown_background()` rather than waiting for it, and that is safe in any context. `drop_off_runtime` is the belt, this is the braces.

That invariant was only proved for `NatsPublisher`. `KafkaPublisher` and `LdapClient` now carry the same two tests: drop inside `block_on`, and drop from a spawned task that outlives the local reference, which is exactly the shape you describe. Replacing `shutdown_background()` with a plain drop turns all six red with `Cannot drop a runtime in a context where blocking is not allowed`, so they assert the invariant rather than restating it. Joining the handles would work too, but it would mean threading a `JoinSet` through every spawn site to defend something the `Drop` impls already guarantee.

**Keep the profile directory in `gateway_binary`** (minor). Declining the change: `current_exe()` returns the test executable *file*, `target/<profile>/deps/<name>-<hash>`. The first `pop()` removes the file name and leaves `deps`, the second removes `deps` and leaves `target/<profile>`. The result is `target/<profile>/barbacane`, which is correct.

The comments caused this. They labelled each `pop()` with the directory it left behind, which reads as though the file name is never removed. They now say what each call drops.

Verified by accident, and it is the better evidence: running these tests against a `target/debug/barbacane` built from a branch without this fix produced `the gateway panicked while shutting down` and `Cannot drop a runtime in a context where blocking is not allowed`. The test found the binary, ran it, and caught the bug. Against this branch's binary both pass.

**Add a regression test for a real HTTPS plugin download** (minor). Already there, in `crates/barbacane-compiler/src/download.rs`, module `runtime_context_tests`: two tests calling `download_plugin` with an `https://` URL from inside a multi-thread runtime, one from `block_on` and one from a spawned task. They exercise the panic precisely, since `reqwest::blocking` refuses on *construction* inside a runtime, before any network I/O. The host is a reserved `.invalid` domain so they do not depend on the network being reachable, and they assert an error rather than a panic. A test against a real reachable host would add network flakiness without testing anything more.

**Also fixed while here**, not raised but adjacent to your first point: both lifecycle tests returned early when the binary or the fixture was missing. `cargo test` renders that as a pass with the reason in captured output, which is how a test silently stops testing. `build_artifact` now returns the reason and both tests fail with it, matching every other test in this crate, which already refuses to run without the binary.

ndreno:
@coderabbitai review

ndreno:
@coderabbitai on the `download.rs` finding from [review 5264473493](https://github.com/barbacane-dev/barbacane/pull/211#pullrequestreview-5264473493), which I missed replying to earlier.

The mechanism is right, and so are the numbers: `std::thread::scope` joins, so the caller's thread is blocked for the whole download, bounded by the client's `timeout(120s)` with a `connect_timeout(30s)`.

Declining the change, because there is no caller where that costs anything.

`download::download_plugin` has exactly one call site, `manifest.rs:325` inside `resolve_url_plugin`, reached only through `resolve_used_plugins`. That path belongs to the manifest-based compile, which is the CLI. `barbacane compile` exists to compile and exit; there is no other task wanting the worker.

The control plane never reaches it. `resolve_used_plugins` appears nowhere in `barbacane-control`: the worker reads each plugin from its own registry with `plugins_repo.get_with_binary`, builds `PluginBundle`s directly, and hands them to `compile`. No URL is resolved and nothing is downloaded.

Wrapping the download alone in `spawn_blocking` would also not work here without restructuring: `resolve_url_plugin` is synchronous, several frames below a synchronous `compile`, so the await would have to be threaded up through the whole call chain to reach an async caller.

The second half of the suggestion is the valuable one, and it is about something else. "Offload the complete synchronous compilation operation" is exactly right for the control plane, which calls the synchronous `barbacane_compiler::compile` straight from an `async fn` at `worker.rs:130` with no `spawn_blocking` anywhere in the crate. Spec parsing, schema compilation, WASM hashing and writing the archive all run on the worker that polled the task, for as long as the project takes. That is a server blocking its executor, and it is much larger than the download. Filed as #216.

PR #212, opened 2026-09-21
fix(compiler): parse AsyncAPI's security scheme types

### LINKED ISSUE #209: AsyncAPI security scheme types are rejected as unknown, so the document fails to parse
## Symptom

The nightly Spec Corpus job fails on a published AsyncAPI example:

```
parse: SchemaError("security scheme 'certs' has unknown type 'X509'")
```

Four corpus checks fail with it: `every_spec_parses`, `schemas_are_self_contained`, `every_local_reference_resolves`, `every_schema_compiles`.

Source: [`streetlights-kafka-asyncapi.yml`](https://raw.githubusercontent.com/asyncapi/spec/master/examples/streetlights-kafka-asyncapi.yml), the AsyncAPI project's own example.

```yaml
components:
  securitySchemes:
    certs:
      type: X509
      description: Download the certificate files from service provider
```

That file also uses `type: scramSha256`.

## Cause

`parse_security_schemes` (`crates/barbacane-compiler/src/spec_parser/parser.rs`) knows the five types OpenAPI defines (`apiKey`, `http`, `oauth2`, `openIdConnect`, `mutualTLS`) and errors on anything else. AsyncAPI defines a larger set, including `X509`, `scramSha256`, `scramSha512`, `gssapi`, `symmetricEncryption`, `asymmetricEncryption`, `httpApiKey`, `plain` and `userPassword`.

The parser applies the OpenAPI list to AsyncAPI documents, so a valid AsyncAPI document is rejected outright. Not just the scheme: the whole spec fails to parse.

## Regression

Introduced in 0.11.0 by the ADR-0033 security model, which made an unknown `type` a hard error rather than an ignored entry. Before that, schemes were not parsed at all and an AsyncAPI document compiled fine.

## What the fix has to decide

An unknown type is only an error because the allowlist wants to know which header carries the credential. For set 3 the question is narrower: does this scheme put a credential in a request header?

For the AsyncAPI types the answer is mostly no. `X509` is a certificate presented at the transport, like `mutualTLS`. `scramSha256`, `scramSha512`, `gssapi` and `plain` are broker mechanisms, not request headers. `httpApiKey` is the AsyncAPI spelling of an `apiKey` in a header and does contribute one. `userPassword` is broker credentials.

So the AsyncAPI types should be recognised and contribute what they actually carry, rather than being rejected. `httpApiKey` maps onto the existing `ApiKey` variant; the rest contribute no header, like `mutualTLS`.

Whether a genuinely unknown type stays a hard error is a separate call. Erroring catches typos, but it also means one unrecognised value makes a whole document uncompilable, which is what happened here.

## Reproduction

```bash
curl -sLO https://raw.githubusercontent.com/asyncapi/spec/master/examples/streetlights-kafka-asyncapi.yml
cargo run -p barbacane -- validate --spec streetlights-kafka-asyncapi.yml
```

The corpus job covers it once fixed; it is what found this.

ndreno:
Closes #209.

## The problem

`parse_security_schemes` matched OpenAPI's five types (`apiKey`, `http`, `oauth2`, `openIdConnect`, `mutualTLS`) and returned a hard error for anything else. AsyncAPI has its own, larger vocabulary, so a single `type: X509` failed the **whole document**, not just that scheme.

The AsyncAPI project's own example is one such document, which is how the nightly Spec Corpus job found it:

```
curl -sLO https://raw.githubusercontent.com/asyncapi/spec/master/examples/streetlights-kafka-asyncapi.yml
cargo run -p barbacane -- validate --spec streetlights-kafka-asyncapi.yml
# security scheme 'certs' has unknown type 'X509'
```

## The fix

Three groups, each mapped by what it means for the request header allowlist:

| AsyncAPI type | Maps to | Header contribution |
|---|---|---|
| `httpApiKey` (name + in) | `SecurityScheme::ApiKey` | the named header, or `cookie` |
| `apiKey` with `in: user\|password` | `SecurityScheme::Transport` | none |
| `X509`, `symmetricEncryption`, `asymmetricEncryption`, `scramSha256`, `scramSha512`, `gssapi`, `plain`, `userPassword` | `SecurityScheme::Transport` | none |

`Transport { kind }` is a new variant keeping the declared type for diagnostics. It sits with `MutualTls` in the allowlist match: a credential the transport or the broker carries reaches no request header.

The `apiKey` arm needed splitting because AsyncAPI reuses the word for a broker credential placed in the connection's user or password field, which has no `name`. The shared arm demanded one and rejected it.

## Verification

- `parses_every_asyncapi_security_scheme_type` covers all twelve. Confirmed it fails without the fix, on both halves: `security scheme 'BrokerUser' of type apiKey has no 'name'` and `security scheme 'Certs' has unknown type 'X509'`.
- `tests/fixtures/asyncapi-security.yaml` puts the shape in the corpus so the PR-time run covers it too.
- The spec above now validates (4 unrelated `x-barbacane-dispatch` warnings).
- Ran the nightly job's full published corpus locally (petstore, stripe, github, box, asana, both AsyncAPI examples, four Kubernetes groups): 7/7 green.
- `cargo clippy --workspace --all-targets -D warnings` and `cargo test --workspace --exclude barbacane-test` clean.

No artifact format change: `ARTIFACT_VERSION` stays at 6.

## Summary by CodeRabbit

* **New Features**
  * Added support for AsyncAPI security schemes, including transport-based credentials, API keys, HTTP API keys, certificates, encryption, and authentication mechanisms.
  * Recognizes credentials carried by the transport or broker without requiring request headers.
* **Bug Fixes**
  * Improved parsing of API key locations, including user and password credentials.
* **Tests**
  * Added coverage for all supported AsyncAPI security scheme types and representative secure broker configurations.


ndreno:
@coderabbitai review

ndreno:
@coderabbitai the finding is correct and reproducible. Confirmed two ways: `servers` is not read anywhere in `parser.rs`, and a document declaring `httpApiKey` in `servers.<name>.security` compiles to `"allowed_request_headers": []`, so the credential header it declares is dropped at ingress.

Tracked separately as #214 rather than folded in here, for three reasons:

1. AsyncAPI 3's `security` is a list of Security Scheme Objects or `$ref`s, not OpenAPI's `{SchemeName: [scopes]}` map, so `parse_security` does not read it and a `$ref` has to be resolved back to the component name.
2. A channel may declare `servers: [...]`, so the applicable requirement is the union over the servers its channel applies to. `channels.<name>.servers` has to be parsed as well.
3. Giving AsyncAPI operations a security requirement starts applying E1057, E1032 and E1072 to them. That is the right outcome and a behaviour change worth landing on its own.

This PR is the prerequisite: without it an AsyncAPI document using any of these scheme types fails to parse at all, which is what the nightly Spec Corpus job caught.

PR #215, opened 2026-09-21
feat(compiler): refuse a security scheme the auth plugin cannot read

### LINKED ISSUE #207: E1032: refuse a security scheme an auth middleware does not implement
## The gap

`E1057` (0.11.0) requires an operation running an authentication middleware to name a security scheme, and requires that scheme to describe a credential the client sends. It does not check that the plugin can actually *read* that kind of credential.

So this compiles cleanly and then rejects every request:

```yaml
x-barbacane-middlewares:
  - name: jwt-auth
    config: { ... }
security:
  - ApiKeyQuery: []
components:
  securitySchemes:
    ApiKeyQuery:
      type: apiKey
      in: query
      name: token
```

`jwt-auth` reads a bearer token from `authorization`. The scheme says the credential is a key in the query string. Both halves are individually valid, the pairing is not, and the first sign is a 401 on every call.

This is a scheme-*type* mismatch rather than a location one, which is why `E1057` cannot see it. ADR-0033 records it as belonging to `E1032`.

## Why it was left open

Catching it means the compiler must know which scheme types each plugin implements. `E1057` deliberately avoids per-plugin knowledge: it keys off `category = "authentication"` in `plugin.toml`, so a third-party auth plugin behaves the same way without the compiler knowing its name. Any fix here has to preserve that.

## Sketch

A plugin declares the scheme types it implements, in `plugin.toml` alongside `category`. Roughly:

| plugin | implements |
|---|---|
| `basic-auth` | `http:basic` |
| `jwt-auth` | `http:bearer`, `oauth2`, `openIdConnect` |
| `oauth2-auth`, `oidc-auth` | `oauth2`, `openIdConnect` |
| `ldap-auth` | `http:basic` |
| `apikey-auth` | `apiKey` |

The compiler then errors when no scheme the operation's requirement names is one the plugin implements.

This also delivers the other half ADR-0033 reserves for `E1032`: a referenced scheme with no matching auth middleware.

Note `E1040` is already taken by `UndeclaredPlugin`, so the scheme-defined-but-never-referenced check needs a free code.

## Prior art in the codebase

- `E1057` and the `category` mechanism: `require_security_requirement` in `crates/barbacane-compiler/src/artifact.rs`
- `E1072`, which warns when a scheme and a plugin's configuration name different credential headers, and is derived from set 4 rather than per-plugin knowledge. Worth copying that approach if it stretches this far.
- ADR-0033, "Authentication requires a security requirement"

ndreno:
Closes #207.

## The gap

`E1057` requires an operation running an authentication plugin to name a security scheme, and requires that scheme to describe a credential the client sends. It never checked that the plugin can read **that kind** of credential. So this compiled cleanly and answered 401 to every request:

```yaml
x-barbacane-middlewares:
  - name: jwt-auth
security:
  - ApiKeyQuery: []
components:
  securitySchemes:
    ApiKeyQuery: { type: apiKey, in: query, name: token }
```

`jwt-auth` reads a bearer token from `authorization`; the scheme says the credential is a key in the query string. Both halves are valid on their own, which is exactly why nothing saw the pairing.

## What the plugin declares

An authentication plugin now lists the scheme types it reads in `plugin.toml`:

```toml
[plugin]
category = "authentication"
implements = ["http:bearer", "oauth2", "openIdConnect"]
```

| plugin | implements |
|---|---|
| `basic-auth`, `ldap-auth` | `http:basic` |
| `jwt-auth`, `oauth2-auth`, `oidc-auth` | `http:bearer`, `oauth2`, `openIdConnect` |
| `apikey-auth` | `apiKey` |

An `http` scheme is qualified by its authentication scheme, since reading Basic credentials is not reading a Bearer token. Matching is case-insensitive, and one match among the schemes an operation names is enough, which is what lets a chain run two auth plugins against an `OR` requirement.

This deviates from the issue's sketch on one row: it gives `oauth2-auth` and `oidc-auth` `http:bearer` as well as `oauth2`/`openIdConnect`. Both read `Authorization: Bearer`, so a document declaring `type: http, scheme: bearer` against either is correct and must not be refused.

**A plugin declaring no list is exempt.** That preserves the property #207 asked for: `E1057` keys off `category` rather than plugin names, so a third-party auth plugin works without the compiler knowing it, and this does the same.

## The other direction: E1033

The issue also asks for the reverse case, a referenced scheme with no matching auth middleware. That lands as a **warning**, not an error: the operation requires a credential and no authentication plugin runs in its chain, so the credential is forwarded and the request reaches the upstream unauthenticated. A warning because the upstream may be the one checking, which is a legitimate topology.

It found three of our own fixtures. `cel.yaml`, `opa-authz.yaml` and `rate-limit.yaml` each declared `security: BasicAuth` at the **root** while only one of their operations ran `basic-auth`, so every other operation claimed to be authenticated and was not. The requirement moves onto the operation that has the plugin. That is 8 warnings turned into 0, and the fixtures now say what they mean.

The ROADMAP line asking for this is removed, and `SPEC-004`/`SPEC-007` are corrected: they described `E1032` as "auth middleware missing for secured route", which is now `E1033`.

## The CI cache bug this turned up

`plugin.toml` and `config-schema.json` are embedded into the `.wasm` at build time, and the compiler prefers the embedded copy over any sidecar file. The plugin cache key hashed `Cargo.toml`, `Cargo.lock`, `*.rs` and the SDK, but **neither of those two files**. A change to either would hit a stale cache and the rebuilt binary would keep the old copy, silently disabling every check that reads them. Both are now in the key.

This predates this PR and would have made `implements` a no-op in CI while passing locally.

## Verification

- Six unit tests on `require_implemented_scheme` and one end-to-end compile test for `E1033`. Disabling the check turns three of the six red plus the `E1033` test; the other three assert acceptance and are there to catch false positives, so they stay green by design.
- End to end against a real `jwt-auth` + `apiKey` document: refused with the message naming both sides. Swapping the scheme to `http: bearer` compiles.
- Every fixture in `tests/fixtures/` compiled: no `E1032`, no `E1033`.
- `cargo test -p barbacane-test --test access_control --test rate_limit` (the suites driving the three edited fixtures): 34 passed.
- `cargo clippy --workspace --all-targets -D warnings` and `cargo test --workspace --exclude barbacane-test` clean.

No artifact format change: `ARTIFACT_VERSION` stays at 6.

The six plugin lockfiles refresh from 0.10.0 to 0.11.0, a side effect of building the auth plugins so the new manifest field is embedded.

## Summary by CodeRabbit

- **New Features**
  - Authentication plugins can declare the credential schemes they support.
  - Validation rejects incompatible plugin and security-scheme pairings with error E1032.
  - Secured operations without authentication middleware produce warning E1033 while forwarding credentials upstream.
  - Security validation covers API keys, HTTP schemes, OAuth2, OpenID Connect, mutual TLS, and broker transports.

- **Documentation**
  - Added guidance and CLI reference entries for authentication validation behavior and related error codes.


ndreno:
Rebased onto main now that #212 has landed, which turned up a bug neither branch could have seen on its own.

#212 added `SecurityScheme::Transport`: AsyncAPI's mechanisms whose credential the transport or the broker carries (`X509`, the SCRAM pair, `gssapi`, `plain`, `userPassword`, and `apiKey` in the connection's user or password field). `scheme_carries_a_credential` returned **true** for all of them, so `E1057` accepted an operation running an authentication plugin whose requirement named only one. No middleware can read a credential off a broker connection, which is precisely what E1057 exists to refuse. `mutualTLS` was the only connection-level scheme when that check was written, and the new variant landed on the wrong side of it.

The two changes touch different lines, so git merged them cleanly and neither branch's CI could have caught it.

Fixed in 6d4a163, with a test covering all four shapes: no request credential, no header contribution, E1057 refuses a requirement naming only one, and the allowlist stays empty. Reverting the fix turns it red.

`scheme_token` gains the variant too, which is what the compiler noticed first (the `implements` match stopped being exhaustive).

ndreno:
@coderabbitai review

ndreno:
@coderabbitai review

ndreno:
@coderabbitai all four addressed in 1eb8cbe. Two were real holes.

**Exclude connection-level schemes from the `implements` contract** (major). Correct, and worse than the example: a plugin declaring only `mutualTLS` satisfied a requirement naming `mutualTLS` and an `apiKey` together, and the same held for AsyncAPI's broker mechanisms. The E1033 side already filtered with `scheme_carries_a_credential`; E1032 did not, which is the asymmetry that let it through. It now filters too, so a connection-level scheme takes no part in the match and the mixed requirement is judged on its request credential alone.

`mutualTLS` leaves the documented token vocabulary in all four places you listed, since no plugin can implement one. `scheme_token` keeps the arm, because the match over `SecurityScheme` has to stay exhaustive, and that exhaustiveness is what caught the `Transport` variant when #212 landed.

Test: a plugin declaring `mutualTLS` against a `{Mtls, KeyHeader}` requirement is now refused, while the plugin reading the key still passes. Reverting the filter turns it red.

**Do not emit E1033 when an applied security alternative is empty** (minor). Correct. `security: [{}, {BasicAuth: []}]` makes authentication optional, so the operation requires nothing and the warning was noise. `required_scheme_tokens` returns an empty set when any applied alternative is empty. Test covers both shapes; reverting turns it red.

**ADR wording** (minor). Fixed. It still described E1032 by the rule it had before this work.

**Move E1032 and E1033 out of the `validate` table** (minor). Right, and the table was wrong well beyond those two rows. `validate` parses specs and resolves nothing: it emits E1000-E1004, E1011 and E1020, so ten of the sixteen rows named checks it never runs, including E1040, E1057, E1071 and E1072. Rather than move two rows into a table that had no home, the catalogue now sits under `barbacane compile`, which is the only command that runs all of it, and `validate` lists exactly what it emits.

ndreno:
@coderabbitai the fifth finding, `SPEC-001-compilation.md:130`, was in the review body rather than inline and I missed it on the first pass. Fixed in 81beb0d.

The collision is real. The compiler emits `E1040` for `UndeclaredPlugin` and does so in three places (`error.rs:42`, `worker.rs:236`, and the CLI reference table), while SPEC-001 section 4.5 assigned the same code to `securitySchemes` defined but never referenced. Two conditions, one code, and only one of them implemented.

`E1040` moves to section 4.3 with the meaning it actually has, since a plugin used but not declared in `barbacane.yaml` is a plugin resolution failure rather than a completeness one. The unreferenced-scheme check takes `E1042`, the next free code, verified unused across `crates/`, `specs/`, `docs/` and `adr/`.

ADR-0033 already recorded the conflict ("`E1040` is already in use for `UndeclaredPlugin`, so the remaining check takes a free code") without resolving it. It now names the code.

PR #217, opened 2026-09-21
fix(control plane): compile off the async runtime

### LINKED ISSUE #216: control plane: the whole synchronous compile runs on a tokio worker
## What happens

`crates/barbacane-control/src/compiler/worker.rs:130` calls the compiler straight from an `async fn`:

```rust
let compile_result =
    barbacane_compiler::compile(&spec_path_refs, &plugin_bundles, &output_path, &options);
```

`barbacane_compiler::compile` is a synchronous `pub fn` (`artifact.rs:390`). Everything it does happens on the runtime worker that polled the task: reading every spec file, parsing them, compiling each JSON Schema validator, hashing every plugin `.wasm`, and writing the gzipped tar.

`spawn_blocking` appears nowhere in `barbacane-control`:

```
$ grep -rn "spawn_blocking" crates/barbacane-control/src/ | wc -l
0
```

## Why it matters here and not in the CLI

The CLI's `run_compile` is also synchronous and also blocks a worker, and that is fine: `barbacane compile` exists to do this one thing and then exit.

The control plane is a server. While one compilation runs, the worker polling it cannot poll anything else on that thread: no other compilation, no API request, no database work. Compilation time scales with the number and size of the specs, so a large project stalls a worker for as long as it takes.

## Fix

Wrap the call in `tokio::task::spawn_blocking` and await the handle. The arguments are paths, bundles and options, so they move into the closure without restructuring anything. The surrounding `async fn` already awaits `tokio::fs` either side of it.

Worth auditing the rest of that worker at the same time for other synchronous calls on the async path.

## Not this

Review on #211 raised the related concern about `download_plugin`, which blocks its caller for up to the client's 120s timeout. That one has no reachable impact: `download::download_plugin` is called only from `manifest.rs:325` on the `resolve_used_plugins` path, and the control plane never uses that path. It builds `PluginBundle`s from its own registry (`plugins_repo.get_with_binary`) and passes them to `compile` directly, so it never resolves a URL and never downloads. The only caller that downloads is the CLI, which has nothing else to run.

ndreno:
Closes #216.

## The problem

`barbacane_compiler::compile` is a synchronous `pub fn` (`artifact.rs:390`), and the compilation worker called it straight from an `async fn`:

```rust
let compile_result =
    barbacane_compiler::compile(&spec_path_refs, &plugin_bundles, &output_path, &options);
```

Everything it does happened on the runtime thread that polled the task: reading and parsing every spec, compiling each JSON Schema validator, hashing every plugin `.wasm`, and writing the gzipped tar. For as long as one compilation took, that thread served no API request, no database work and no other compilation.

`spawn_blocking` appeared nowhere in the crate:

```
$ grep -rn "spawn_blocking" crates/barbacane-control/src/ | wc -l
0
```

The CLI has the same shape and it is fine there: `barbacane compile` exists to do this one thing and exit. The control plane is a server.

## The change

The call moves to the blocking pool. Two details worth noting:

- The inputs are not used after the call, so they **move** into the closure rather than being cloned. That matters: `PluginBundle` carries the plugin's WASM bytes, and cloning the bundle list would copy every binary. Only `output_path` is cloned, a `PathBuf`, since the caller reads the artifact back from it.
- Awaiting the handle turns a panic inside the compiler into a failure for that one compilation instead of an unwinding task.

## What is deliberately left

The SHA-256 of the finished artifact stays on the async path. It is a single pass over a few megabytes, milliseconds rather than the order of a compile, and wrapping it would add noise without changing anything measurable.

## Verification

`cargo clippy --workspace --all-targets -D warnings` and `cargo test --workspace --exclude barbacane-test` are clean.

**There is no test, and I want to be straight about that.** `barbacane-control` has no `tests/` directory and the worker has no harness; exercising this path needs a Postgres-backed one that does not exist. Building that harness is worth doing and is a larger piece of work than this fix. What is verified here is that the code compiles with the inputs moved and the result type unchanged.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved control-plane responsiveness during compilation by preventing long-running compilation work from blocking other runtime tasks.
  - Compilation operations involving specification processing, validator builds, plugin verification, and archive creation now run without interrupting normal worker activity.


ndreno:
@coderabbitai review

ndreno:
@coderabbitai review

ndreno:
@coderabbitai both correct, fixed in 74d6e4e. The first is a regression the previous commit introduced.

**Persist failures from `spawn_blocking`.** Right, and worse than a lost error. `process_compilation` had already *claimed* the row before reaching the compile, so returning the `JoinError` through `?` left it sitting in `compiling` with nothing to move it on, since `run_worker` only logs what the function returns. Previously a panic unwound the worker loop, which was a different failure but at least a visible one; converting it to an error without recording it traded a loud break for a silent stuck job.

The join result is now matched explicitly: a failed task calls `mark_failed` with the join error as the message before returning, so the compilation ends in a state a caller can observe.

**Move specification parsing to the blocking pool.** Also right, and it is the same finding as the original issue at a call site I missed. `resolve_project_plugins` runs before `compile` reaches its `spawn_blocking`, and `parse_spec` is YAML plus a full traversal of the document, so a large spec blocks the worker just as the compile did. Reading stays asynchronous; the parse and `extract_plugin_names` now run in `spawn_blocking`.

Still no test, for the reason in the PR description: `barbacane-control` has no `tests/` directory and exercising this path needs a Postgres-backed harness that does not exist. Both changes are verified by compilation and by reading. The missing harness is worth its own issue, and the `mark_failed` path you found is a good argument for it.

PR #218, opened 2026-09-21
test(harness): surface what the gateway logged when a test fails

### LINKED ISSUE #213: test harness: a failing integration test throws away the gateway's own explanation
## Observed

CI run [35578434286, Integration Tests (shard 1)](https://github.com/barbacane-dev/barbacane/actions/runs/35578434286/job/106267622972):

```
---- test_opa_unreachable_returns_problem_json stdout ----
assertion `left == right` failed
  left: 500
 right: 503
```

`test_opa_unreachable_returns_503` sits directly above it in the same file, drives the same spec (`tests/fixtures/opa-authz.yaml`), issues the same `GET /opa-protected`, and passed in the same run. So the 500 is not the code under test disagreeing with the assertion. Something went wrong inside that one gateway, and the CI log says nothing about what.

## Why nothing says what

Two places hold the answer and both are dropped on the floor.

**1. The gateway's log.** `crates/barbacane-test/src/gateway.rs:217-218` pipes the child's stdout and stderr:

```rust
.stdout(Stdio::piped())
.stderr(Stdio::piped());
```

They are read in exactly one place: the startup path, when the child has already exited (`gateway.rs:298-314`). A gateway that starts fine and then logs an error keeps it in an undrained pipe until `Drop` kills the process and closes it (`gateway.rs:460-466`).

Every 500 out of the middleware path is logged before it is returned:

```rust
tracing::error!(error = %e, "middleware chain execution failed");
return Err(self.dev_error_response(format_args!("middleware chain error: {}", e)));
```

That line names the cause. It has never reached a CI log.

**2. The response body.** The harness runs the gateway with `--dev` (`gateway.rs:208`), and `dev_error_response` puts the cause in the problem detail in dev mode. The test asserts the status first, so the body it was about to read is dropped with the failed assertion.

## Fix

- Drain the child's stdout and stderr on a thread from spawn, keep the tail in the `TestGateway`, and print it when an assertion fails. A `Drop` that prints on panic, or an accessor the test can call, either works.
- Assert status with a helper that includes the response body in the message, so a dev-mode gateway explains itself without any extra plumbing.

Neither changes what is tested. They change a red run from "a number was wrong" into "here is the error the gateway logged".

## On the failure itself

Not diagnosed, deliberately: the evidence above is what exists, and it is not enough to name a cause with confidence. Worth noting for whoever picks it up that the 500 can only have come from the gateway, since `plugins/opa-authz/src/lib.rs` answers 503 on every `OpaError` path, and that `max_execution_ms` defaults to 100 (`crates/barbacane-wasm/src/limits.rs:30`) and is a wall-clock epoch deadline that also covers instantiation and `init` on a gateway's first request. A loaded runner stalling one thread past that budget is a candidate, not a finding.

ndreno:
Closes #213.

## The problem

This was the failure that prompted the issue:

```
---- test_opa_unreachable_returns_problem_json stdout ----
assertion `left == right` failed
  left: 500
 right: 503
```

A status and nothing else. The gateway logs the cause of every error it answers with, including the `middleware chain execution failed` line behind that 500, and none of it had ever reached a CI log.

`TestGateway` piped stdout and stderr, then read them in exactly one place: the startup path, and only when the child had **already exited**. A gateway that started fine and then logged an error kept it in an undrained pipe until `Drop` killed the process and closed it.

## The change

Both streams are drained from spawn, one reader thread each, into a bounded tail on the `TestGateway`.

- **`Drop` prints it when the thread is panicking.** Every existing assertion in the suite gains the explanation without being rewritten.
- **`TestGateway::log()`** exposes the same output to a test that wants to assert on it (`.text()`, `.contains()`).
- **The startup path** reads the drained tail instead of the pipe it no longer owns, so a gateway that dies during boot still reports what it said on the way out.

Draining removes a hazard as well as adding output: a piped stream nothing reads fills its buffer and blocks the writer, so a talkative gateway could stall on its own logs.

The tail is capped at 400 lines. `--dev` logs every dropped header on every request and would otherwise outrun anything reading it; the recent lines are the ones that explain a failure.

## The other half of the issue

The issue also asked for a status assertion carrying the response body. That already exists: `assert_status` panics with `Response body:\n{body}`. The failing test simply used a bare `assert_eq!(resp.status(), 503)` instead. The `Drop` print covers those cases too, which is why this does not touch the tests themselves.

## Verification

`crates/barbacane-test/tests/harness.rs`, two tests:

- the harness captures the gateway's startup output
- an error response is explained by the log: `opa-authz.yaml` points the plugin at a dead port, and after the request the log names the failure

Removing the drain turns both red, so they assert the fix rather than describe it.

Existing suites unaffected: `access_control` (28), `lifecycle` (20), `validation` (2), `process_lifecycle` (14) all pass. `cargo clippy --workspace --all-targets -D warnings` is clean.

## Summary by CodeRabbit

- **New Features**
  - Test gateway output is captured continuously while the gateway runs.
  - Added access to captured logs, including full-text retrieval and bounded search.
  - Startup failures now display relevant gateway output for easier diagnosis.

- **Bug Fixes**
  - Improved gateway log handling during failures, including connection errors.

- **Documentation**
  - Added an unreleased changelog entry describing the updated test gateway logging behavior.


ndreno:
@coderabbitai review

ndreno:
@coderabbitai both correct, fixed in 28ad4db. The first is the race this PR exists to remove, still present in its own implementation.

**Join log readers before rendering captured output.** `drain` dropped each reader's handle, so a reader could be part way through the output when `Drop` or the startup path called `text()`. The lines that explain a failure are the last ones written, which makes them exactly the ones most likely to be lost, so the change would have been unreliable at the one moment it matters.

`GatewayLog` now keeps the handles and gains `join_readers`. Your note about ordering was the load-bearing part: a reader ends at EOF, EOF arrives when the child closes its pipes, so the child is killed and reaped first and only then are the readers joined. Joining earlier would wait on a process still running. `Drop` and the startup-failure path both follow that sequence now, and the 50ms sleep that stood in for it is gone.

**Wait for each expected log record with a bounded deadline.** Same problem from the test side. The fixed 200ms sleep is a race whichever way it is tuned: too short and it flakes, too long and it slows every run without ever being provably enough. `GatewayLog::wait_for(needle, timeout)` polls with a deadline and both tests use it, including the immediate read after startup that you flagged at line 26.

Verified: harness tests pass, and `access_control` (28), `lifecycle` (20) and `process_lifecycle` (2) are unaffected. `cargo clippy --workspace --all-targets -D warnings` is clean.

PR #219, opened 2026-09-21
fix(compiler): carry AsyncAPI server security to the operation

### LINKED ISSUE #214: AsyncAPI server security does not reach the request header allowlist
## The gap

In AsyncAPI 3 a security requirement is declared on the **server**, not on the operation. `parse_asyncapi_channels` sets every generated operation's `security` to `None`, and `servers` is not read anywhere in the parser at all. `operation_header_allowlist` reads only operation and root `security`, so a credential header an AsyncAPI document declares is never admitted and is dropped at ingress.

## Reproduction

```yaml
asyncapi: "3.0.0"
info: { title: Server Security, version: "1.0.0" }
servers:
  gateway:
    host: broker.example.com
    protocol: ws
    security:
      - $ref: '#/components/securitySchemes/apiKeyHeader'
channels:
  events:
    address: /events
operations:
  publish:
    action: send
    channel: { $ref: '#/channels/events' }
    x-barbacane-dispatch: { name: mock, config: { status: 200 } }
components:
  securitySchemes:
    apiKeyHeader:
      type: httpApiKey
      name: X-API-Key
      in: header
```

Compiles cleanly, and the operation in `routes.json` carries:

```json
"allowed_request_headers": []
```

`X-API-Key` is the credential the document declares, and the gateway drops it.

## Why it is not simply "parse servers.security"

Three things differ from the OpenAPI path:

1. **Shape.** AsyncAPI 3's `security` is a list of Security Scheme Objects or `$ref`s to them, not OpenAPI's `{SchemeName: [scopes]}` map. `parse_security` cannot read it. A `$ref` has to be mapped back to the component name to look the scheme up.

2. **Which server applies.** A channel may list `servers: [...]`; when absent it applies to every server. So an operation's effective requirement is the union over the servers its channel applies to, which means `channels.<name>.servers` has to be parsed too.

3. **Most schemes contribute nothing.** `X509`, `scramSha256` and the rest are transport credentials carrying no request header (`SecurityScheme::Transport`), so only `httpApiKey`, `http` and the OAuth pair change the allowlist. That bounds the blast radius: a document has to use one of those on a server for anything to be dropped.

## Scope

Found by review on #212, which taught the parser AsyncAPI's security scheme *types*. That change is what makes these schemes representable at all; wiring the server requirement through to the operation is the separate piece and is not in it.

Once operations carry the requirement, `E1057`, `E1032` and `E1072` start applying to AsyncAPI operations as they already do to OpenAPI ones, which is the right outcome but is a behaviour change worth landing deliberately.

ndreno:
Closes #214. Found by CodeRabbit's review of #212.

## The problem

AsyncAPI declares the security requirement on the **server**, not the operation. The parser read neither `servers` nor `channels.<name>.servers`, and set every generated operation's `security` to `None`:

```rust
// AsyncAPI carries no OpenAPI security requirement.
security: None,
```

So a document declaring a credential got it dropped at ingress. Before:

```json
"allowed_request_headers": []
```

After, for the same document:

```json
"allowed_request_headers": ["x-api-key"]
```

The document said the connection was authenticated and the gateway removed the one thing that made it so.

## Resolution

Following AsyncAPI's own rules:

- A server's `security` entries are **alternatives** (satisfying one authorizes the connection), so each becomes its own requirement object.
- A channel narrowed with `servers: [$ref]` inherits from those servers alone.
- A channel naming no servers is reachable on every server, so it inherits from all of them.
- An entry is normally a `$ref` into `components.securitySchemes`. It may also be an **inline scheme object**, which has no name of its own. The allowlist is keyed on names, so an inline scheme is registered under a synthesized one rather than silently contributing nothing.
- A `$ref` to a scheme the document does not define is refused: it resolves to nothing, and admitting no header while claiming to be authenticated is the bug this fixes.

`parse_security_scheme` is split out of `parse_security_schemes` so `components` and inline server schemes read the same vocabulary.

## A false positive this exposed

E1072 compared the plugin's configured header against **each** scheme header in turn, warning about every one the plugin did not read. That was rare while requirements came from OpenAPI documents, where an operation usually names a single credential scheme. It is the common case for an AsyncAPI channel reached through several servers, and my own new fixture tripped it immediately:

```
warning[E1072]: the security scheme says 'x-internal-key' carries the credential,
but 'apikey-auth' is configured to read 'x-api-key'
```

Requirements are alternatives, so the plugin agrees with the document as soon as it reads one of them. It now warns only when the plugin matches none.

## Consequence worth stating plainly

E1057, E1032 and E1072 now apply to AsyncAPI operations as they already did to OpenAPI ones. That is the point of giving them a requirement at all, but it is a behaviour change: an AsyncAPI document running an authentication plugin on a server with no security now fails E1057, where before it compiled and rejected every request at runtime.

## Verification

- Six parser tests: inheritance, channel narrowing, alternatives, inline registration, absent security, dangling `$ref`. Removing the inheritance turns three red.
- One allowlist test that scheme headers are alternatives.
- `tests/fixtures/asyncapi-server-security.yaml` puts the whole path in the PR-time corpus: two servers with different credentials, one channel narrowed and one not, `apikey-auth` reading each.
- Reproduction from the issue now yields `["x-api-key"]`.
- The AsyncAPI project's `streetlights-kafka-asyncapi.yml` still validates: its server schemes are `scramSha256` and `X509`, which the connection carries, so they contribute no header.
- Nightly corpus run locally over the published specs: 7/7.
- `cargo clippy --workspace --all-targets -D warnings` and `cargo test --workspace --exclude barbacane-test` clean. Every repository fixture compiles with no new E1032/E1033/E1057/E1072.

## Summary by CodeRabbit

- **Bug Fixes**
  - AsyncAPI operations now inherit security requirements from their servers and channel-specific server settings.
  - Credentials declared for AsyncAPI servers now reach upstream requests.
  - Improved validation for undefined or invalid security scheme references.
  - Header mismatch warnings now correctly handle alternative credential headers without unnecessary alerts.
  - Added support for inline server security schemes and anonymous access where applicable.

- **Documentation**
  - Clarified how AsyncAPI, HTTP, OAuth2, and OpenID Connect security schemes provide request authorization headers.


ndreno:
@coderabbitai review

ndreno:
@coderabbitai both correct, fixed in 1f5bccf.

**Preserve the anonymous alternative only for mixed server reachability.** Right, and the consequence was the wrong direction: the operation read as authenticated on every route to it, including the one that needs no credential. `parse_server_security` skipped servers with absent or empty `security`, so a channel reachable through a secured and an unsecured server inherited only the secured requirement.

Every server is now recorded, an unsecured one with an empty list, and `server_security_requirement` appends the empty `SecurityRequirement` when the applicable servers include both. Your last sentence was the part worth heeding: recording the empty list alone changes nothing, because the requirement builder collects scheme names and an empty list contributes none. A channel whose applicable servers are all unsecured still yields `None`, so nothing gains a requirement it did not have.

This composes with the E1033 change from #215: an empty alternative means authentication is optional, so the warning correctly does not fire on such an operation while the credential header stays admitted.

**Reject invalid and dangling channel server references.** Right, and this one mattered more than it looks. `servers: [$ref: '#/servers/interal']` narrowed the channel to a server that does not exist, matched nothing in `server_security`, and returned `None`, leaving the operation unconstrained and the declared credential dropped at ingress. That is precisely the bug this branch exists to fix, reachable again through a typo, and silently. `parse_channel_servers` now returns `ParseError` and requires each entry to be a `$ref` into `#/servers/` naming a defined server.

Tests for both: a channel on a secured and an unsecured server keeps the anonymous alternative while a channel narrowed to the secured one does not, and a channel naming `interal` is refused. Reverting either fix turns the corresponding test red.

The AsyncAPI project's `streetlights-kafka-asyncapi.yml` still validates, and every repository fixture compiles with no new E1032/E1033/E1057/E1072.

PR #222, opened 2026-09-22
perf(compiler): hold schema definitions once per document

### LINKED ISSUE #221: Spec Corpus fails nightly: parsing stripe.json peaks at 5.7 GB and the runner is reclaimed
The nightly Spec Corpus job has failed every day. There are two separate causes; one is already fixed.

| Date | Cause |
|---|---|
| 18-20 Sep | `security scheme 'certs' has unknown type 'X509'` — #209, fixed by #212 |
| 21-22 Sep | Exit 143, `The runner has received a shutdown signal` |

The second is not a test failure. The runner is reclaimed part way through.

## Why the runner goes away

The job's peak resident set is far past what a hosted runner can give it:

```
$ BARBACANE_CORPUS_DIR=... /usr/bin/time -l cargo test -p barbacane-compiler --test corpus
test result: ok. 7 passed
      154.97 real
          8789114880  maximum resident set size      # 8.79 GB
```

Bounding parallelism does not help. It is worse:

```
$ ... --test-threads=2
      169.67 real
         10846781440  maximum resident set size      # 10.85 GB
```

So it is not seven tests holding seven copies. Individual specs are the cost.

## One spec dominates

Peak RSS for `barbacane validate` on a single spec, which parses and nothing more:

| Spec | File | Peak RSS | Ratio |
|---|---|---|---|
| **stripe.json** | **7.7 MB** | **5.67 GB** | **~735x** |
| github.json | 12 MB | 0.19 GB | ~16x |
| k8s-core.json | 2.0 MB | 0.36 GB | ~180x |
| petstore.json | small | 0.02 GB | - |

A 7.7 MB document taking 5.67 GB, while a 12 MB one takes 0.19 GB, is not "large spec" behaviour.

## It is not the number of references

```
stripe.json: $ref count = 3919
github.json: $ref count = 10625
```

GitHub has nearly three times as many references and uses thirty times less memory. What differs is what sits behind the hot ones:

```
stripe.json: most-referenced -> #/components/schemas/error         (594 times)
github.json: most-referenced -> #/components/responses/not_found   (601 times)
```

`not_found` is 128 bytes and effectively a leaf. Stripe's `error` is 0.2 KB flat, so flat inlining explains nothing either, which means the cost is **transitive**: `error` refers on into Stripe's deeply interlinked schema graph (`error` → `api_errors` → `customer` → `subscription` → ...), and each of the 594 sites expands that whole closure again.

As an aside on scale: a small Python script computing that transitive closure, with cycle guarding and a depth cap of 12, did not terminate in five minutes.

## What this suggests

`$ref` resolution appears to inline by value with no sharing between sites, so a modestly sized component that reaches a large subgraph is duplicated once per reference. Sharing resolved subtrees (`Arc`, or resolving to an index rather than a copy) would collapse this, and would help every large real-world document rather than only the corpus job.

## Why it matters beyond CI

`validate` alone needs ~6 GB for a published Stripe spec. Anyone pointing Barbacane at a document of that shape hits the same wall on their own machine or in their own pipeline, and the compile path does strictly more work than `validate`.

## Reproduction

```
curl -fsSL -o stripe.json https://raw.githubusercontent.com/stripe/openapi/master/openapi/spec3.json
/usr/bin/time -l barbacane validate --spec stripe.json
```

ndreno:
Closes #221.

## The measurement that started it

Parsing Stripe's published document needed **5.67 GB for 7.7 MB of JSON**. GitHub's, at 12 MB, needed 0.19 GB. So not document size.

Nor reference count: github.json has **three times more** `$ref`s (10,625 vs 3,919) and used thirty times less memory.

## The cause

Resolution already shares. Within one schema a definition is resolved once into `$defs` and pointed at from every use, which the code says and explains. But `resolve_schema` started a **fresh set per schema**, and there are six call sites: request body, each response, each parameter.

Stripe is the worst case for that. `#/components/schemas/error` is referenced 594 times, exactly once per operation, and reaches most of a 1454-component graph. Nearly every operation carried a copy of nearly the whole document.

Truncating the spec shows it is linear, not combinatorial:

| paths | peak RSS |
|---|---|
| 10 | 0.23 GB |
| 50 | 0.80 GB |
| 200 | 3.03 GB |
| 419 | 5.67 GB |

## The change

One pool per document, threaded through parsing, stored once on `routes.json`.

Two documents compiled into one artifact can each name a definition `error`, so merging renames a collision and rewrites that document's pointers.

The data plane attaches definitions before compiling a validator, **and attaches only the ones the schema reaches**. That last part is not an optimisation, it is the difference between fixing this and moving it: attaching the whole pool to every schema measured 844s against 155s before, where reaching only what is needed measures 98s.

`ARTIFACT_VERSION` 6 → 7. A v6 artifact carries its definitions inside each schema and still validates, because a schema that already has `$defs` is left alone. A v7 artifact needs a data plane that knows about the pool, which is what the version bump enforces.

## Results

Parsing `stripe.json`:

| | before | after |
|---|---|---|
| peak RSS | 5.67 GB | **0.09 GB** |
| time | - | 0.62s |

The corpus suite over the published specs, which is what the nightly job runs:

| | before | after |
|---|---|---|
| peak RSS | 8.79 GB | **1.74 GB** |
| time | 155s | **98s** |

Faster as well as smaller, which was not the goal but follows from copying less.

## Verification

- End to end against a running gateway, including a **transitive** reference (`Thing` → `Nested`), which is the shape that blew up: valid body 200, wrong nested type 400 with the right message, missing required field 400.
- A regression test builds a document where N operations all reference the same three-deep shared schema and asserts the pool stays at 3 entries whether N is 2 or 200, and that no schema carries its own `$defs`. Reverting to per-schema pools fails it with `pool: None`.
- The two corpus invariants that asserted per-schema self-containment now assert it at document level, which is the contract that replaced it.
- `cargo clippy --workspace --all-targets -D warnings` and `cargo test --workspace --exclude barbacane-test` clean; `validation`, `routing` and `access_control` integration suites pass.

## Scope

This fixes the compiler. The data plane still builds one validator per schema, each holding its own compiled closure, which was true before and is unchanged. Whether that needs addressing is a separate question and not claimed here.

## Summary by CodeRabbit

- **Changed**
  - Reduced memory usage when compiling large API documents by storing shared schema definitions once instead of duplicating them.
  - Updated schema reference handling so validators continue resolving shared definitions correctly at runtime.
  - Large documents, such as Stripe’s published specification, now require significantly less memory to process.

- **Compatibility**
  - Compiled artifacts use a new format version. Existing artifacts must be recompiled before loading them with the updated gateway.


ndreno:
@coderabbitai both correct, fixed in 3cef52b. The first is a data-corruption bug in code I wrote in this PR.

**Deduped pool entries get mutated in place.** Right, and the reasoning goes one step further than the symptom: the dedup itself was unsound. Two byte-identical bodies are not interchangeable, because each carries references resolved against its own document. `Wrapper` can be identical in both specs while the `Inner` it points at differs, so sharing one entry is already wrong before any rewrite touches it. The in-place mutation is how that unsoundness became visible.

So the fix is not to narrow the rewrite loop but to drop the dedup: a name already taken is renamed whether or not the bodies match. Every spec then owns the entries it contributes, and the rewrite can only reach its own. The cost is one duplicate entry per genuinely-identical shared name across specs, which is nothing next to correctness, and nothing at all for the single-spec case.

Test added: two documents share a `Wrapper` name and differ in the `Inner` beneath it, and each operation is followed through the pool to the `Inner` its own document declared. Restoring the dedup fails it with `operation /string resolved to the other document's Inner`, which is your scenario exactly.

**Escaped pointer segment against unescaped rename keys.** Also right, and in both directions as you noted: reading without unescaping, and writing the new name without escaping. Both now go through the same `pointer_unescape` / `pointer_escape` the parser uses, which I made crate-visible rather than duplicating.

Worth saying that this is the second time this exact mismatch appeared in this PR. The first was in `collect_reachable_defs`, caught by the existing `pointer_escaped_definition_name_resolves` conformance test. That test does not reach the artifact merge path, which is why this one needed you.

Verified: `cargo clippy --workspace --all-targets -D warnings` and the workspace suite clean, corpus 9/9.

PR #224, opened 2026-09-22
feat(compiler): refuse a plugin config the plugin's own schema rejects

### LINKED ISSUE #220: E1023 is specified but never implemented: plugin config is not validated against its schema
## What happens

A plugin configuration that its own `config-schema.json` forbids compiles cleanly and fails at runtime.

```yaml
x-barbacane-dispatch:
  name: ai-proxy
  config:
    targets:
      cloud:
        provider: anthropic
        model: claude-sonnet-5     # not a TargetConfig property
        api_key: "env://KEY"
    default_target: cloud
```

```
$ barbacane compile --spec bad-config.yaml --manifest barbacane.yaml -o bad.bca
compiled 1 spec(s) to bad.bca (1 routes, 1 plugin(s) bundled)
```

Then, at the first request through that operation:

```
{"detail":"failed to get plugin instance: plugin initialization failed:
  plugin ai-proxy init returned 1","status":500}
```

Every request through the dispatcher fails. Nothing said so at compile time.

## The schema already forbids it

```
TargetConfig: additionalProperties=False,
  props=['provider', 'api_key', 'base_url', 'auth', 'allow', 'deny']
```

`model` is absent and `additionalProperties` is `false`, so a validator would reject this immediately. The plugin does reject it, in `AiProxy`'s deserializer, and the test that covers it calls itself "the runtime safety net for the schema gap":

```rust
#[test]
fn config_rejects_legacy_model_on_target() {
    // Same migration check applied to nested targets — the runtime safety
    // net for the schema gap that vacuum doesn't recurse into yet.
```

The gap is wider than the comment suggests. It is not only vacuum.

## The cause

`E1023` is documented in `specs/SPEC-001-compilation.md:114` ("Plugin config does not match the plugin's declared JSON Schema"), with a sample message at line 245. It is implemented nowhere:

```
$ grep -rn "E1023" crates/ --include='*.rs'
$
```

No occurrences. The compiler never validates a middleware or dispatcher `config` against the plugin's schema.

## Why it is a small fix

Everything needed is already in place in `compile_inner`:

- The schemas are loaded and indexed. `artifact.rs:738` builds `plugin_schemas: HashMap<&str, &serde_json::Value>` from each bundle's `config_schema`, and the existing header-annotation and `writeOnly` passes already walk them.
- `jsonschema` is a dependency of `barbacane-compiler` and is already used to build validators for request and response schemas.
- The middleware and dispatcher configs are already in hand at the point where `E1057`, `E1072` and the allowlist are computed.

So it is: compile the plugin's schema once per plugin, validate each middleware and dispatcher `config` against it, and report `E1023` with the plugin name and operation, matching the message already in the spec.

## Where it should not fire

A bundle without its schema, which is what `E1071` already warns about (control-plane compiles, URL-sourced plugins). Those skip the check the way they skip the header and security ones.

## Impact

This was found upgrading a real deployment from 0.9.0 to 0.11.0. `model` on a target was valid in 0.9.0 and removed by ADR-0030, so an artifact that compiled clean took the gateway's AI routes down with a 500 on every request. The compiler held the schema that would have caught it.

ndreno:
Closes #220.

## The gap

`E1023` is specified in `SPEC-001` ("Plugin config does not match the plugin's declared JSON Schema") with a sample message at line 245, and implemented nowhere:

```
$ grep -rn "E1023" crates/ --include='*.rs'
$
```

The compiler already held each plugin's schema — it uses it for header annotations and for finding plaintext secrets — and never validated a configuration against it.

## What that cost

`ai-proxy` stopped accepting `model` on a target in ADR-0030: the client names the model and the gateway routes on it. A deployment carried the old form past a clean compile, the plugin then refused to initialise, and every request through that dispatcher answered:

```
{"detail":"failed to get plugin instance: plugin initialization failed:
  plugin ai-proxy init returned 1","status":500}
```

Nothing said so until the gateway was serving it. With this change:

```
error: compilation failed: E1023: invalid config for plugin 'ai-proxy':
  POST /chat in 'bad-config.yaml': Additional properties are not allowed ('model' was unexpected)
```

## What is deliberately not checked

- **An operation that names no `config`.** The plugin applies its own defaults and a schema does not always restate them, so validating an absent config as `{}` would reject working specs. This was a false positive in my first attempt, caught by the fixture sweep.
- **A plugin bundled without its schema**, which `E1071` already warns about — control-plane compiles and URL-sourced plugins.

Schemas are compiled once per artifact rather than once per operation.

## It found two real faults immediately

On its first run against this repository's own fixtures:

- `asyncapi-server-security.yaml` gave `apikey-auth` a **`consumer`** field it has no concept of (the entry takes `key`, `id`, `name`, `scopes`). My own fixture, wrong since I wrote it.
- `observability.yaml` asked `mock` for **`delay_ms: 50`** so an endpoint would "violate SLO" against a 10ms target. `mock` implements no delay — `grep -n delay plugins/mock/src/lib.rs` is empty — so that endpoint has always responded immediately and the description described something that never happened. No test covered it, so nothing was silently passing, but the fixture was lying.

Both fixed here.

## Verification

- Reproduces the real failure and rejects it, naming the offending field.
- A valid config for the same plugin still compiles, so the check is not simply refusing `ai-proxy`.
- Disabling the check makes the test fail with the legacy config compiling cleanly.
- Every fixture in `tests/fixtures/` compiles with no `E1023`, as do the three specs from the deployment this came from.
- `cargo clippy --workspace --all-targets -D warnings` and `cargo test --workspace --exclude barbacane-test` clean; corpus 11/11; `validation` and `routing` integration suites pass.

`jsonschema` moves from a dev-dependency to a dependency, since the compiler uses it now and not only the tests.


ndreno:
Both focus areas from the local review were worth answering.

**Silently dropping a schema that fails to build.** Fixed in deacd16. It was skipped without a word, so that plugin's configuration went unchecked and nothing said so, which is the same failure this whole check exists to end. It now warns under `E1071`, which already covers a plugin arriving without the files these checks read. I have not found a way to test the path without a plugin that publishes a malformed schema, so it is verified by reading rather than by a test.

**Validating an absent `config` against `{}` instead of skipping.** Keeping the skip, deliberately. Validating `{}` is what I wrote first and it rejected three of this repository's own fixtures: plugins apply defaults their schemas do not restate, so an operation that names no config is not the same as one that names an empty one. The suggestion is right that a missing required field then goes uncaught; the alternative rejects specs that work today, which is worse. A plugin that truly requires configuration still refuses at init, as it did before.

The ticket-compliance check against #220 was a nice surprise, and accurate.

PR #226, opened 2026-09-23
Build the plugins the corpus checks compile against

ndreno:
Closes the nightly failure in [run 35845392465](https://github.com/barbacane-dev/barbacane/actions/runs/35845392465).

## What broke

Four of the eleven checks in `crates/barbacane-compiler/tests/corpus.rs` compile a fixture through a plugin, and the Spec Corpus workflow never built one. Three return early when the `.wasm` is missing, so they reported success without running anything. The fourth, added alongside the shared `$defs` pool, calls `compile` directly:

```
compile: PluginResolution("failed to read plugin 'mock' from
.../plugins/mock/mock.wasm: No such file or directory (os error 2)")
```

Reproduced locally by moving `plugins/mock/mock.wasm` aside: same panic, same line. Restoring it turns the test green again, which is why it passed in review.

## Why nobody noticed

Nothing else runs this test target. The pull request suite runs `cargo test --workspace --lib --bins`, which builds no integration test target, and `scripts/coverage.sh` covers the workspace and `barbacane-test`. The nightly is the only place these eleven checks execute, and four of them were not executing.

The workflow header claimed "the same checks run on every pull request against the repository's own fixtures". That is not true, and the comment is removed.

## The fix

`mock`, `rate-limit`, `basic-auth` and `ai-proxy` are built for wasm32 before the checks run, each artifact copied to the name the fixtures reference. The crate name is not the directory name, so `barbacane_mock_dispatcher.wasm` becomes `mock.wasm`, the way the Makefile does it.

The plugin cache gets its own key rather than reusing the pull request workflow's. This builds four plugins, and saving that under the shared key would hand a later `make plugins` a cache hit for the twenty-nine it had not built.

`zizmor --min-severity=medium` is clean.

## Still open

The corpus target running nowhere but the nightly is worth changing on its own: a pull request that breaks one of these eleven checks finds out the next morning, on `main`. Adding `cargo test -p barbacane-compiler --test corpus` to a pull request job that already has plugins built would close that, and against repository fixtures rather than fetched specs it should be quick. Left out of this change deliberately.


ndreno:
Dispatched on this branch: [run 35857919371](https://github.com/barbacane-dev/barbacane/actions/runs/35857919371), green.

```
  mock.wasm (176K)
  rate-limit.wasm (228K)
  basic-auth.wasm (236K)
  ai-proxy.wasm (1.3M)
...
test result: ok. 11 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 195.04s
```

No `skipping:` lines, so all four plugin-dependent checks ran for the first time. 195s against 155s for the failing run, the difference being four cold wasm builds; later runs hit the new cache.

PR #227, opened 2026-09-23
Let the gateway pick its own test port instead of naming one

### LINKED ISSUE #204: test harness: the gateway port is released before the child binds it, so tests collide
`find_available_port` in `crates/barbacane-test/src/gateway.rs` asks the OS for a free port, then gives it up before anything uses it:

```rust
let listener = std::net::TcpListener::bind("127.0.0.1:0")?;
let port = listener.local_addr()?.port();
drop(listener);
Ok(port)
```

The number is then passed to a gateway child process as `--listen 127.0.0.1:<port>`. Between the `drop` and the child binding, the port belongs to nobody, so another test running in parallel can be handed the same one.

When that happens a request reaches **another test's gateway**, which has a different spec and does not know the path, so it answers **404** where the test expects a real status. That is the signature to recognise: a 404 in place of an expected 2xx, in a test that otherwise passes.

## Observed

`Coverage (full)` on main at 7852186 failed:

```
---- test_cors_credentials_header stdout ----
assertion `left == right` failed
  left: 404
 right: 204
```

Not a regression: the same code passed at bd22aa1, a direct descendant containing every change in the failing commit, and at fa23f56 before it. It reddened a main-branch run for no reason, which is the cost.

## Fix

Stop round-tripping through a closed socket. Either:

1. **Keep the listener and pass the descriptor to the child**, so the port is never unowned. Cleanest, and removes the race rather than shrinking it.
2. **Retry on bind failure**: have the child report that it could not bind, and let the harness pick another port and restart. Tolerates the race instead of preventing it.
3. **Partition the range per test binary** from an environment variable, so binaries cannot collide even if the window stays.

The first is the real fix. The others only narrow the window.

## Why it is worth doing

The suite runs sharded and in parallel in CI, which is exactly the condition that produces the collision, and a spurious red on main is most costly at the moment it is least welcome, such as during a release.

ndreno:
Closes #204.

## The race

`find_available_port` bound a socket to ask the OS for a free port, closed it, then passed the number to the gateway child:

```rust
let listener = std::net::TcpListener::bind("127.0.0.1:0")?;
let port = listener.local_addr()?.port();
drop(listener);
Ok(port)
```

Between the `drop` and the child's bind the port belongs to nobody. A concurrent test handed the same one sends its request to the other test's gateway, which has a different spec and answers 404 for a path it has never heard of. The suite runs sharded and parallel in CI, which is exactly the condition that produces it, and the cost is a spurious red on main.

## The fix

The child binds `127.0.0.1:0` for both the data and admin ports and reports what it got. No port is ever unowned, and the harness never names one. That removes the race rather than narrowing it, the same property as the descriptor-passing option in the issue, without teaching the gateway to inherit a file descriptor.

Two data plane changes make it possible, both correct on their own terms:

- `serve` and `dev` logged the address they were *asked* for, which reads `127.0.0.1:0` when the OS is choosing. They log `listener.local_addr()` now.
- The admin API bound inside its spawned task, so its port was not known when the line announcing it was written. It binds first, and `start_admin_server` takes the listener.

`GatewayLog` gains `wait_for_line`, which returns the matching line rather than just reporting that one exists; `wait_for` is now a thin wrapper over it.

## Evidence

- `the_harness_never_names_a_port` guards the constant; `reads_the_port_off_a_startup_line` covers the parser including the `https` form. Both fail against a mutated source.
- Reverting the `serve` log line to print the requested address **and rebuilding the binary** fails all six admin tests with `gateway did not become ready in time`. That is what makes the data plane change load-bearing rather than cosmetic. Worth noting for anyone testing this area: `cargo test -p barbacane-test` does not rebuild the gateway binary, so a stale one will happily pass.
- Locally green: 25 harness unit tests, plus `admin`, `harness`, `process_lifecycle`, `routing` and `proxy` (38 integration tests). The full sharded suite runs here in CI.

## Note on scope

`find_available_port` is gone, with no remaining callers anywhere in `crates/` or `plugins/`.


ndreno:
Two updates.

**Shard 3 failed, my fault.** Five CEL-heavy tests died with `gateway did not report a listen port`, all still in startup when the wait expired. I had put a 30-second port wait in front of a health check the code already documents as needing 120 for exactly this case, two CEL gateways cold-booting on a shared runner. The new wait became the binding constraint and reported an unfinished boot as a failed one.

Both waits now share `STARTUP_TIMEOUT`, so neither can be the tighter of the two and the reason for the number lives in one place. A missing admin line is now an error rather than a port of 0, which was only reachable when the admin API never announced itself and would have surfaced later as a connection failure with nothing pointing at the cause.

**Scope correction.** This PR fixes `find_available_port` in the shared harness, which is what #204 describes. Two tests bypass the harness and spawn a child themselves, each with a private copy of the same pattern, so the race survives in both: `security/artifact_integrity.rs` and `security/authz.rs`. Filed as #229 rather than widened into this one, since the `authz` case spawns `barbacane-control` and needs the control plane to log its bound address the way the data plane now does.

Worth flagging from #229: in `artifact_integrity.rs` the collision is not a spurious 404 but a spurious **pass-to-fail on a security assertion**. It asserts a tampered artifact is rejected, and a collision with another test's healthy gateway answers the health probe, so it reports "loaded" and fails claiming tampering was accepted.

ndreno:
Note for merge order: #230 builds on this and had to be retargeted to `main` to get any CI, because `ci.yml` only triggers on pull requests based on `main`. It therefore carries these two commits as well until this merges, at which point its diff shrinks to its own. Merging this first is the tidier order, but either works.

PR #230, opened 2026-09-23
One way to learn a child's port, used by all three spawners

### LINKED ISSUE #229: security tests spawn their own child on a port they release first, the same race as #204
#227 fixes `find_available_port` in the shared harness. Two tests bypass the harness and spawn a child process themselves, each with a private copy of the same pattern, so the race survives in both.

## `crates/barbacane-test/tests/security/artifact_integrity.rs`

```rust
fn free_port() -> u16 {
    let l = TcpListener::bind("127.0.0.1:0").expect("bind ephemeral");
    let p = l.local_addr().expect("local addr").port();
    drop(l);
    p
}
```

Used for both `--listen` and `--admin-bind` on a `barbacane serve` child. This one is worse than a spurious 404, because of what the test asserts. `gateway_loads_artifact` returns whether the gateway became healthy, and the callers assert it did **not**, the point being that a tampered artifact must be rejected. A collision with another test's healthy gateway answers the health probe, so the function reports "loaded" and the test fails claiming tampering was accepted. A security test that goes red for a reason unrelated to security is a bad failure to have to triage.

Fixable the same way #227 fixes the harness: `serve` now logs its bound address, so the child can be given `127.0.0.1:0` and its port read from stderr. It needs its own stderr drain because it deliberately does not use `TestGateway`, whose `start` returns `Err` for exactly the case this test is asserting.

## `crates/barbacane-test/tests/security/authz.rs`

```rust
fn free_port() -> Option<u16> {
    let listener = std::net::TcpListener::bind("127.0.0.1:0").ok()?;
    let port = listener.local_addr().ok()?.port();
    drop(listener);
    Some(port)
}
```

Used for `--listen` on a `barbacane-control serve` child. The same fix needs the control plane to log its bound address the way the data plane now does, so this one is a slightly wider change.

## Not affected

Mock servers in `plugins.rs`, `http_client.rs` and `ldap_client.rs` also bind `127.0.0.1:0`, but they keep the listener and serve on it. The port is never unowned, so there is nothing to fix there.

ndreno:
Closes #229. **Stacked on #227**, so the base is that branch; GitHub retargets this to `main` when #227 merges.

## Why a helper rather than a third fix

#227 fixed `find_available_port` in the shared harness. Two tests spawn a child of their own and each carried a private copy of the same pattern, which is the actual root cause. So this adds the abstraction that was missing instead of patching two more copies.

`LoggedChild` owns spawning with drained pipes and reading back an announced port. `TestGateway::start`, `security/artifact_integrity.rs` and `security/authz.rs` all use it. Both `free_port` helpers are deleted.

## The one that mattered

`artifact_integrity.rs` asserts a tampered artifact is **rejected**. A port collision with another test's healthy gateway answers the health probe, so the helper reported "loaded" and the test failed claiming tampering had been accepted. A security test going red for a reason unrelated to security is an expensive thing to triage, especially during a release.

`authz.rs` spawns `barbacane-control`, which logged the address it was *asked* for rather than the one it bound, the same bug the data plane had. It logs `local_addr()` now. CI's Security Suite provides Postgres, so that path is exercised there even though it skips locally.

## Two things the shared version gets right

- **A child that exits without announcing ends the wait immediately.** Three of the four `artifact_integrity` tests expect the gateway to refuse the artifact. Waiting out the full two-minute timeout for each would have added eight minutes to the suite.
- **A reported port of 0 is treated as no port.** That is exactly what a child prints when it echoes the address it was asked for, which is the bug #227 fixed in the gateway. Taking it at face value produces a connection failure a long way from its cause.

Ports are read anchored on the loopback host rather than the last colon, because the control plane logs through `tracing`, which appends its own fields after the message.

## Coverage

13 tests, 1.0s. Every announcement format in use including the control plane's two, a port of 0, a non-loopback address, the u16 boundary, the needle choosing between two announcements, a slow announcement, one behind a backlog larger than the log keeps, and both ways of giving up with timing assertions separating "the child died" from "the limit expired".

Checked against mutated sources. Trusting a port of 0, reverting the parser to splitting on the last colon, and removing the early-exit branch each turn specific tests red; the early-exit mutation takes the full 60 seconds, which is the point of it.

**One honest gap.** `announced_port` calls `join_readers` before its last look on the exit path, so a reader still holding the announcement is waited for. I could not make a test that fails without it: the 50 ms poll interval already gives the reader enough time, even behind a 4000-line backlog. The call stays because a loaded runner has no such guarantee, but it is defensive and uncovered rather than tested. The backlog test was renamed to what it does verify.

Locally green: 36 lib tests, 21 security (4 pre-existing `#[ignore]`), plus `admin` and `harness`.

## Summary by CodeRabbit

* **New Features**
  * Gateway and admin servers can bind to an OS-assigned port when configured with port `0`, allowing the operating system to choose an available port.
* **Bug Fixes**
  * Startup messages report the actual bound address, including an OS-assigned port. If the gateway cannot retrieve its bound address, the requested address is shown instead.
  * If the admin API cannot bind to its requested address, the gateway continues serving traffic and reports that the admin API is unavailable.


ndreno:
Retargeted from `fix/204-test-port-race` to `main`.

`ci.yml` triggers on `pull_request: branches: [main]`, so a pull request based on any other branch gets **no CI at all**. Stacked as it was, this had exactly one check, CodeRabbit's "review skipped" status, and nothing had built or run. Unverified code is worse than a noisy diff.

The consequence is that this now carries #227's two commits as well, since they are not on `main` yet. Merging #227 first shrinks this diff back to its own commit automatically. Either order works; CI covers the whole of it now.

Worth knowing separately: that trigger makes stacked pull requests unverifiable in this repository. Not something to change during a release, but it is a trap.

ndreno:
Both review findings were correct and both were mine. Fixed in 419b7de and 37a52aa. Recording the reasoning here since both threads auto-resolved and went outdated on the push.

## 1. An admin bind failure became fatal (419b7de)

Binding the admin listener before spawning its task, so the port could be logged, also moved the failure into the startup path, where I returned `ExitCode::from(1)`. A port conflict on the admin API used to log `Admin server failed` from inside the task and leave the gateway serving. I turned a degradation into an outage, in `serve` as well as `dev`, and had not noticed.

Both paths now report the conflict and skip the admin server. `/health`, `/metrics` and `/provenance` go missing; traffic does not.

**The `drop_off_runtime` half of that finding does not apply.** No path in `run_dev` calls it, including the pre-existing listen-bind failure at line 4281 that predates this PR. Adding it to only the branch I introduced would have made it the odd one out. Whether `run_dev` ought to follow `run_serve` here is a fair question, but it is not something this PR introduced and widening the change during a release is the wrong trade. Moot for this branch anyway, since neither admin branch returns early now.

## 2. The test proved too little, and raced (37a52aa)

`wait_for_port` returns on the first successful TCP connect, and the listening socket stays connectable while the process is on its way out, so it could pass against a gateway that was dying. The earlier mutation run was caught only because the process happened to die before the first poll. That is timing, not a guarantee.

It now waits for `admin API unavailable` to be reported, then asks `/__barbacane/health` for a 2xx. Ordering the two is the part that matters: the conflict has already been handled by the time the request goes out, so a gateway answering on its way down cannot satisfy it.

Reading `log.text()` before `kill_and_reap` raced the reader threads for the very line the assertion was about to check. `kill_and_reap` joins them, so the read follows it now. This repository fixed exactly that bug in #218, and I reintroduced it in a test written to check someone else's work.

Raw HTTP/1.1 over a socket rather than pulling in a runtime for one request, keeping the test synchronous like the two beside it.

## Verification

Re-checked against the mutation. Restoring the fatal branch still fails the test, now on the report assertion first, since that path prints `failed to bind admin API to` rather than `admin API unavailable`; the response assertion fails too, so it is covered twice. The tightened test also runs faster, 1.7s against 3.4s.

ndreno:
Rebased onto `main` after #227 merged.

This carried #227's two commits under their original SHAs. The repository squash-merges, so #227 landed as one new commit (`ab769ae`) and those two would have collided with the later commits here that rewrite the same regions, notably the admin bind block. I checked that rather than assumed it: simulating the squash and then merging this branch conflicted in `gateway.rs` and `main.rs`.

Rebasing `--onto main` from `0bb7bf3` dropped the two now upstream and replayed the remaining three cleanly. #227's branch and `main` hold identical code, differing only in `.coderabbit.yaml` and the deleted workflow because #227 predated #228, so there was nothing to reconcile. `admin.rs` has correctly dropped out of this diff, having arrived with #227.

What remains here: the shared `LoggedChild` helper and its 13 tests, the two security tests moved onto it, the control plane logging its bound address, the admin-bind regression fix, and the tightened lifecycle test.

Green locally against the rebased branch: 36 lib, 21 security (4 pre-existing `#[ignore]`), 3 process lifecycle, 6 admin, 2 harness.

ndreno:
Correct, and the fix belongs on the helper rather than the call site. 8ac23fc.

`Child` does not kill on drop and `LoggedChild` had no impl of its own, so any early return holding one leaves the process running for the rest of the test binary. `TestControlPlane::boot` returns exactly that way, and it does so on the **skip** path, which is the one nobody looks at: a control plane and its database connection left behind while the test reports as skipped.

Rather than patching `authz.rs`, `Drop` now kills and reaps, so every caller is covered including ones not written yet. That is the same reasoning as the helper itself: this bug existed in three copies because each spawner solved it privately.

One wrinkle worth recording. The child had to become an `Option`. `into_parts` hands ownership to a caller that keeps the process running, and a field cannot be moved out of a type that implements `Drop`; taking it also makes the drop of what remains a no-op, which it must be. Joining the reader threads there would wait for an EOF that a still-running child never sends, so the naive version hangs rather than leaks.

Two tests, asking the OS rather than trusting the code:

- `dropping_a_child_stops_it` captures the pid, drops, and asserts `kill -0` fails. Removing the impl fails it with `pid N outlived the LoggedChild that spawned it`.
- `handing_the_child_over_leaves_it_running` asserts the reverse for `into_parts`, since killing a process `TestGateway` still needs would be the opposite bug.

Honest limit: I could not construct a mutation that breaks the second without restructuring the type, so it documents that property rather than being mutation-verified. The first is verified.

Local: 38 lib, 21 security (4 pre-existing `#[ignore]`), 3 process lifecycle, 6 admin, 2 harness.

